PR #3, opened 2023-12-13
Feature/add user registration and log in endpoints

saulRM10:
**Summary**

1. Created User, Role and Group Models. Included migration create the tables in the database. May notice that the User migration contains explicit associations between the Group and Role tables as `groupId` and `roleId` are foreign keys in the user table 
2. Created controllers for each model, with simple create functionality 
3. User model contains functionality to hash and salt user's passwords when creating a new password using the `Bcrypt` package. Also uses it to compare passwords when user logs in. This hashing and compare passwords logic is in the user model file but its used in the `config/passport.js` file. 
4. Cleaned up `app.js` by using router, thus created `router/modelName.routes.js` file system. 

**Testing :**
To create a new user, role , and or  group using postman you would do the following: 

1. if in main, run `git pull`
2. `git checkout  feature/add-user-registration`
3. run migration so you get the new User, Group and Role tables !!!!!: `npx sequelize-cli db:migrate`
4. Double check your database and make sure new migrations ran successfully and you have the new tables 
5. Go to post man and  set to send a `post` request
6. `http://localhost:80/[model-name]/[end-point-that-can-be-found-in-the-route-files-in-the-route-folder` 
7. go to the **Body** tab 
8. select `x-www-form-urlencoded`
9.  set up key value pairs. Reference the model to get the names of the keys. Note if you want to create a new user like in the screenshot provided there needs to be an existing group and role because you need to include a `groupId` and `roleId` 
Screenshot 2023-12-13 at 3 45 49 PM

### DIFF
--- a/app.js
+++ b/app.js
@@ -5,6 +5,8 @@ const teamController = require("./controllers/team.controller")
 const UserController = require("./controllers/user.controller")
 const GroupController = require("./controllers/group.controller")
 const RoleController = require("./controllers/role.controller")
+const session = require('express-session');
+const passport = require('./config/passport');
 
 const bodyParser = require('body-parser');
 
@@ -16,10 +18,27 @@ app.use(bodyParser.urlencoded({ extended: true }));
 app.set('port', process.env.PORT || 80)
 app.use(express.static(path.join(__dirname, 'client', 'build')));
 
+app.use(session({
+  secret: 'your-secret-key',
+  resave: false,
+  saveUninitialized: false,
+}));
+
+app.use(passport.initialize());
+app.use(passport.session());
+
 function init() {
   app.get('*', function(req, res) {
     res.sendFile('index.html', {root: path.join(__dirname, 'client', 'build')});
   });
+
+  app.post('/login', passport.authenticate('local'), (req, res) => {
+    const welcomeMessage = req.user
+    ? `Welcome, ${req.user.firstName} ${req.user.lastName}`
+    : null;
+  res.json({ message: welcomeMessage });
+  });
+  
 
   app.post('/create/:name', teamController.createTeam);
   app.post('/user', UserController.registerUser);
--- a/app.js
+++ b/app.js
@@ -7,7 +7,7 @@ const GroupController = require("./controllers/group.controller")
 const RoleController = require("./controllers/role.controller")
 const session = require('express-session');
 const passport = require('./config/passport');
-
+const UserRoutes = require('./routes/user.routes');
 const bodyParser = require('body-parser');
 
 const http = require('http');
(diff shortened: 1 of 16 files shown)


config/passport.js
```diff
@@ -0,0 +1,34 @@
...
+});
+
+passport.deserializeUser(async (id, done) => {
+  try {
+    const user = await User.findByPk(id);
+    done(null, user);
+  } catch (err) {
+    done(err);
+  }
+});
```
abanuelo, line 32: Nice job in supporting session based middleware for passport. When adding middleware, there is an inherent tradeoff we need to make. [It is best explained here](https://www.passportjs.org/concepts/authentication/sessions/).

I recommend the following strategy:
```
passport.serializeUser((user, done) => {
    done(null, {
      id: user.id,
      email: user.email,
      firstName: user.firstName,
      lastName: user.lastName,
      groupId: user.groupId,
      roleId: user.roleId,
    );
});

passport.deserializeUser(async (user, done) => {
  try {
    done(null, user);
  } catch (err) {
    done(err);
  }
});
```

This way we don't have to hit the database on every request to for the deserialization and we can simply return the user information we captured in the serialization inside our session.

app.js
```diff
@@ -2,16 +2,40 @@ const path = require('path');
...
 const app = express();
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({ extended: true }));
 app.set('port', process.env.PORT || 80)
 app.use(express.static(path.join(__dirname, 'client', 'build')));
 
+app.use(session({
+  secret: 'your-secret-key',
+  resave: false,
+  saveUninitialized: false,
```
abanuelo, line 26: I think we need to the add: ` cookie: { secure: true }` to the session.

abanuelo (changes requested):
Great start! I did some debugging following the testing strategy and found some bugs:

1. As part of the testing strategy, users need to first run `npm install` before starting the server. Minor nit but just to keep things documented you may want to update the description in the pull request. 
2. Running `POST` `http://localhost:80/role/create` will generate issues when you explicitly do not define `role_type` in postman causing the entire server to fail. Inside the Role controller check to see if `req.body.role_type` is defined before running the sequelize `create` call. Please repeat this for `/group/create`.
3. If you run `POST` `http://localhost:80/user/register` with for a user who has already been registered (ie their email already exists in the User table), the server will also crash. Please add some checks inside the User controller file to first run  a `User.find` a user with that email being passed in. Also check for any entries that cannot be null in the table (ie firstName, lastName, groupId, roleId, etc).

I have added some screenshots for reference:
Screenshot 2023-12-13 at 8 02 00 PM

abanuelo:
@saulRM10 can we close this out? Just wanting to confirm you have addressed these comments into the codebase via another pr #13

PR #13, opened 2023-12-19
Feature/create user log in and registration endpoints

saulRM10:
**### WIP WORKING ON ADDING TESTS** 
**Summary**

1. Created User, Role and Group Models. Included migration create the tables in the database. May notice that the User migration contains explicit associations between the Group and Role tables as `groupId` and `roleId` are foreign keys in the user table 
2. Created controllers for each model, with simple create functionality 
3. User model contains functionality to hash and salt user's passwords when creating a new password using the `Bcrypt` package. Also uses it to compare passwords when user logs in. This hashing and compare passwords logic is in the user model file but its used in the `config/passport.js` file. 
4. Cleaned up `app.js` by using router, thus created `router/modelName.routes.js` file system. 
5.  Updates: Using `validate()` to validate data against the model's constraints !! And all wrapped in the `try` `catch` so if there is an error it won't crash the app

Manual Testing:

1.  if in `main` branch, `git pull`
2.  switch onto the branch `git checkout feature/craete-user-log-in-n-registration `
3. run `npm install`
4. run migration so you get the new User, Group and Role tables !!!!!:  If in root directory then `cd server` as there is where the migration folder is located in and then run `npx sequelize-cli db:migrate`
5. Double check your database and make sure new migrations ran successfully and you have the new tables 
6. Go to post man and  set to send a `post` request
7. `http://localhost:80/[model-name]/[end-point-that-can-be-found-in-the-route-files-in-the-route-folder` 
8. go to the **Body** tab 
9. select `x-www-form-urlencoded`
10.  set up key value pairs. Reference the model to get the names of the keys. Note if you want to create a new user like in the screenshot provided there needs to be an existing group and role because you need to include a `groupId` and `roleId` 
Screenshot 2023-12-13 at 3 45 49 PM


abanuelo (approved):
LGTM! Great work on adding model validation before create requests and explicitly using `router.post` inside the router files. Also great work on adding the necessary info for users in the passport serializeUser function! Also tested that frontend calls to your backend routes work! Great work! 

Honestly if you wanted to drop in test cases in another PR that is cool too and just call it `test/craete-user-login-in-n-registration`, up to you! That way you don't have so much in a single PR 👍 
Screenshot 2023-12-19 at 11 38 33 AM

saulRM10:
> LGTM! Great work on adding model validation before create requests and explicitly using `router.post` inside the router files. Also great work on adding the necessary info for users in the passport serializeUser function! Also tested that frontend calls to your backend routes work! Great work!
> 
> Honestly if you wanted to drop in test cases in another PR that is cool too and just call it `test/craete-user-login-in-n-registration`, up to you! That way you don't have so much in a single PR 👍 Screenshot 2023-12-19 at 11 38 33 AM

I am still working on the tests, but ill go ahead and create a separate pr so @Jen-464  can get access to these endpoints and connect them to the log in page.

PR #15, opened 2024-01-31
Tests for user log in and registration

saulRM10:
**Summary** 
- test for user , group and role controllers 
- test for group and role routes 
- TODO test for the `login` route, hard to test the route because it is using passport.js middleware  `passport.authenticate()`  , having a hard time mocking in the tests 
- Fix: fixed the session issue, it required me to update the way i was calling passport middleware

### DIFF
--- a/server/routes/user.routes.js
+++ b/server/routes/user.routes.js
@@ -3,9 +3,20 @@
 const express = require("express");
 const router = express.Router();
 const UserController = require("../controllers/user.controller");
+const passport = require("passport");
+
+router.get("/loginReactPageHere", (req, res) => {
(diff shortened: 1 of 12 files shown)


server/routes/user.routes.js
```diff
@@ -3,9 +3,20 @@
...
+router.get("/loginReactPageHere", (req, res) => {
+  res.json({ message: "Invalid email or password, try again" });
+});
 
 router.post("/register", UserController.registerUser);
 
-router.post("/login", UserController.logInUser);
+router.post(
+  "/login",
+  passport.authenticate("local", {
```
saulRM10, line 16: call the passport authentication middleware and if successful it will continue to the controller function

PR #26, opened 2024-03-25
feat: add error handler middleware

villaleo:
This change introduces an error handler middleware function.

The error handler middleware streamlines the error handling process. It will enable our errors returned by endpoints to remain consistent. Errors will be in the form of
```json
{
    "message": "An error occurred.",
    "stack": "Error: An error occurred\n at /Users/..."
}
```
The `message` property contains the reason for the error and the `stack` property displays the error's stack trace.
> For safety reasons, the `stack` property will have a value of `[hidden]` when in the production environment.

## Summary by CodeRabbit

- **Refactor**
	- Positioned the `errorHandler` middleware as the last middleware in the Express app setup.
	- Introduced a middleware function `errorHandler` to handle errors by sending an error response with a message and stack trace, hiding the stack trace in the production environment.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -12,6 +12,7 @@ const GroupRoutes = require("./routes/group.routes");
 const RoleRoutes = require("./routes/role.routes");
 const UserRoutes = require("./routes/user.routes");
 const PlayerRoutes = require("./routes/player.routes");
+const { errorHandler } = require("./middlewares");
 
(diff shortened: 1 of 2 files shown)


saulRM10 (approved):
Looks good, thanks for the update of the middleware file

villaleo (commented):
Done!

PR #28, opened 2024-03-27
Create division routes and controller

humbertodiazp:
https://cascarita.atlassian.net/browse/C1-39

## Summary by CodeRabbit

- **New Features**
    - Introduced Division management functionalities within the application, including creating, updating, retrieving, and deleting divisions.
    - Added new server routes for division-related operations.

- **Tests**
    - Implemented comprehensive test cases for division-related routes to ensure functionality covers all specified scenarios, including error handling and data validation.

- **Refactor**
    - Updated relationships in data models to align with new division functionalities, ensuring proper data integrity and association between divisions, groups, and sessions.


abanuelo (changes requested):
Great work so far @humbertodiazp I really like how you have created a folder underneath `/server/__tests__/` called `controllerTests` and 'routeTests'. Since `controllerTests` and `routeTests` are both sub directories underneath the '__tests__', I think we don't need to add the `Tests` suffix to the folder name. Feel free to rename to just the `controller` and `route` folders.

server/routes/division.routes.js
```diff
@@ -5,10 +5,8 @@ const DivisionController = require('../controllers/division.controller');
 const router = express.Router();
 
 router.post('/', DivisionController.create);
-router.put('/:id', DivisionController.update);
-// router.put('/archive/:groupId', DivisionController.archiveByGroupId);
+router.patch('/:id', DivisionController.update);
 router.get('/group/:groupId', DivisionController.getByGroupId);
```
JesusGo115, line 9: This route should be consistent with the other routes. Don't include the /group and just keep the /:groupId

humbertodiazp: Thanks! Updated the routes and related test for consistency.

humbertodiazp:
> Great work so far @humbertodiazp I really like how you have created a folder underneath `/server/__tests__/` called `controllerTests` and 'routeTests'. Since `controllerTests` and `routeTests` are both sub directories underneath the '**tests**', I think we don't need to add the `Tests` suffix to the folder name. Feel free to rename to just the `controller` and `route` folders.


Hey Armando! Sorry I missed your message. Thank you for the feedback. I agree that we don't necessarily need to add the suffix, however, I didn't actually create those folders/sub-folders so not sure I cant make that change. But if we all agree on this, I will go ahead and change it.

server/controllers/division.controller.js
```diff
@@ -0,0 +1,100 @@
...
+            if (!group) {
+                return res.status(400).json({ error: 'Group does not exist' });
+            }
+
+            const isUnique = await isDivisionNameUnique(group_id, name);
+            if (!isUnique) {
+                return res.status(400).json({ error: 'Division name is already taken.' });
+            }
+
+            const division = await Division.create({ group_id, name });
```
saulRM10, line 25: in your model you have a validation object for checking length, to use those validate objects in the model you need to call `modelName.validate(newDivision)`  reference `GroupController.createGroup()` atm it looks like it will allow a user to submit an empty , really short or really long division name

server/controllers/division.controller.js
```diff
@@ -0,0 +1,100 @@
...
+            const { id } = req.params;
+            const { name } = req.body;
+
+            const division = await Division.findByPk(id);
+            if (!division) {
+                return res.status(404).json({ error: 'Division not found' });
+            }
+
+            const group_id = division.group_id;
+            const isUnique = await isDivisionNameUnique(group_id, name);
```
saulRM10, line 45: i notice this gets repeated couple times:

`           const division = await Division.findByPk(id);
            if (!division) {
                return res.status(404).json({ error: 'Division not found' });
            }`
            
            i think that logic could be implemented in you `isDivisionNameUnique` or make a validate parameter for group_id in the model , saying it must exist

server/controllers/division.controller.js
```diff
@@ -0,0 +1,100 @@
...
+            if (!isUnique) {
+                return res.status(400).json({ error: 'Division name is already taken.' });
+            }
+
+            await division.update({ name });
```
saulRM10, line 50: here you can call the `validate()` to ensure it does not let the user update the division name with an empty name, really short or really long name. Reference group controller update function

server/__tests__/routeTests/division.routes.test.js
```diff
@@ -0,0 +1,134 @@
...
+
+    expect(response.status).toBe(404);
+    expect(response.body).toMatchObject({
+      message: "No divisions found for the given group ID",
+    });
+  });
+
+  // ------------------- Create Tests ----------------
+
+  it("should handle POST /", async () => {
```
saulRM10, line 52: add a test case for passing in a really short to really long name, like you specified in the model min 2 chars, max 50 chars

saulRM10 (changes requested):
left some comments, good job Humberto. Add another test case and address my comments  in your controller and we can merge this pr !

PR #37, opened 2024-04-09
fix!: decouple season from group

villaleo:
- Updates `models/season.js` and `models/group.js` to remove the references between `Group` and `Season`.
- Creates a migration to remove the `group_id` column from `Season`, adding a `league_id` column instead and supplying a default value.


villaleo:
It's exactly the same as before. Huge PR with changes that I didn't make.

PR #47, opened 2024-04-30
feat!: Session Cookie Login with CSRF and XSS protection measures

abanuelo:
# IMPORTANT!!!

### What are the changes?
This PR introduces some breaking changes especially within the `server/app.js` file. In particular, the way we define backend `/api/` routes has transitioned from `router.use('/user', UserController)` to `app.use('/api/user', UserController)` as we wish to apply the session cookie to all base routes (frontend + backend), not just those exclusively prefixed with `/api` (which was implicitly defined with `app.use("/api", router)`). 
### Why are the changes needed?
This makes it so that when users navigate to say frontend routes like `/leagues` or `/users`, if the session cookie is expired, it will redirect them back to the `/login` page to restrict their access to sensitive information on the frontend!

## Details

**This PR includes:**
- Session Cookie with an arbitrary expiry (ie 10 seconds for now) to illustrate how users get revoked of their login sessions
- Use of the [`csrf-csrf`](https://www.npmjs.com/package/csrf-csrf) library as the [express`csurf`](https://github.com/expressjs/csurf) library has some concerning security vulnerability with csrf token exposure. This other library addresses these concerns in light of those vulnerabilities
- We enable the session cookie as `http-only`, protecting it from XSS javascript attacks from the frontend

## When do I add CSRF protection to routes?
CSRF protection should be added to routes where sensitive operations are performed via HTTP POST, PUT, PATCH, or DELETE requests. These are typically actions that modify data or state on the server-side. Note, GET's are ignored by default.

## How do we add the Double Submit Cookie Pattern for CSRF protection?
Once authenticated, we can add the following middleware directly to the route you wish to protect: 

```
// define the middleware
const { generateToken, doubleCsrfProtection } = require("../csrf");

// use the middleware doubleCsrfProtection
router.get("/user", doubleCsrfProtection, (req, res) => {
  if (req.isAuthenticated()) {
    res.status(200).json({ user: req.user });
  } else {
    res.status(401).json({ error: "Unauthorized" });
  }
});
```

# Demo

https://github.com/cascaritaco/cascarita/assets/32311654/1f56b649-80e8-4e56-9fcf-b586d0452960

## Summary by CodeRabbit

- **New Features**
  - Introduced a new authentication system, including sign-in, sign-up, and logout functionalities.
  - Added CSRF token handling for enhanced security.
  - Implemented navigation updates to redirect users based on authentication status.

- **Enhancements**
  - Updated the layout structure to include `TopNav` and `SideNav` within a unified `Layout` component.
  - Simplified several pages by removing unnecessary layout dependencies.

- **Bug Fixes**
  - Corrected user login response status from 201 to 200.

- **Documentation**
  - Added detailed comments and updated descriptions across various components and middleware for better clarity.

- **Refactor**
  - Consolidated navigation components into a single layout.
  - Enhanced code consistency with added semicolons and reorganized imports.

- **Style**
  - Implemented new CSS styles for the login page to improve user interface design.


client/index.tsx
```diff
@@ -9,11 +9,9 @@ const root = ReactDOM.createRoot(
   document.getElementById("root") as HTMLElement
 );
 root.render(
-  <React.StrictMode>
```
villaleo, line 12: Is there a reason why we want to stop using strict mode?

server/__tests__/controllerTests/user.controller.test.js
```diff
@@ -201,8 +201,7 @@ describe("logInUser", () => {
 
     await UserController.logInUser(req, res);
 
-    expect(res.json).toHaveBeenCalledWith({ message: "Welcome, Leo Messi" });
-    expect(res.status).toHaveBeenCalledWith(201);
+    expect(res.status).toHaveBeenCalledWith(200);
```
villaleo, line 204: This test name says it should return a 201 for an authenticated user.

server/app.js
```diff
@@ -1,14 +1,46 @@
...
 
-const bodyParser = require("body-parser");
-const csrf = require("csurf");
+const app = express();
+
+app.set("port", process.env.PORT || 80);
+app.use(express.static(path.join(__dirname, "../dist")));
+
+const sessionMiddleware = session({
+  secret: "secret", //TODO: replace with valid secret
```
villaleo, line 18: Let's create an environment variable and reference it here
```
secret: process.env.SESSION_SECRET,
```

server/csrf.js
```diff
@@ -0,0 +1,6 @@
+const { doubleCsrf } = require("csrf-csrf");
+const { generateToken, doubleCsrfProtection } = doubleCsrf({
+  getSecret: () => "secret",
```
villaleo, line 3: We can use the secret here, too
```
process.env.SESSION_SECRET
```

villaleo (approved):
Armando, really really good stuff! Give yourself a pat on the back. I left 3 comments-- really small issues, but overall I think it's perfect. Thank you for tackling login/logout and sessions! I learned a lot.

client/index.tsx
abanuelo: Correct, if we have the `React.StrictMode` active it actually causes the `useEffect()` [React hook to trigger twice](https://sentry.io/answers/react-useeffect-running-twice/#:~:text=This%20occurs%20because%20since%20React,developers%20uncover%20bugs%20during%20testing.); If you feel like there are other alternatives here please let me know! I just found it annoying!

server/__tests__/controllerTests/user.controller.test.js
abanuelo: Ack! Will make those changes accordingly!

server/app.js
abanuelo: Great catch! Also as a best practice there are two different secrets we need to create: 
1. one to encrypt the session cookie (ie `process.env.SESSION_SECRET`)
2. one to encrypt the CSRF token (see server/csrf.js for details); I will create another secret called `process.env.CSRF_SECRET`

These two need to be different to ensure higher levels of security. Ack! Will make those changes accordingly!

server/csrf.js
abanuelo: Please see the comment above, this needs to be a different secret value just for us to have higher levels of security in our application. I will create a new env variable called `process.env.CSRF_SECRET` for reference.

PR #54, opened 2024-05-14
feat: Adding Multilingual Support

abanuelo:
### Description

Given that clients need to platform to be multilingual, we have added multilingual support. Note that for this demo as well users can reload the page and they will stick with their preferred language preference.

#### Demo

https://github.com/cascaritaco/cascarita/assets/32311654/73db869e-d87d-49f4-9155-51ae5ef12d58

### Checklist


saulRM10 (approved):
Pulled and tested the branch, looks good !

PR #57, opened 2024-05-16
Chore/update group return response

saulRM10:
### Description

- Updated the return object of the group controller to just return the group object , no need for the `success` boolean 
- lowercased all the error messages 
- updated tests

### Checklist


villaleo (approved):
That's perfect! Thanks for updating those response messages.

PR #60, opened 2024-05-24
Chore: Update Env with db connection variables

saulRM10:
### Description

- got rid of `config.json` , it was used to get db connection variables.
- Moved them all to the `env` file and deleted  `config.json` 
- updated `index.js` to use the env file to set up db connection instead of  `config.json` 
 
**NOTE**
- devs you need to update your env with the changes i added to the `env.example` !!!!

### Checklist


saulRM10:
turn out we need the `config/cofig.json ` based on the sequalize docs

PR #62, opened 2024-05-26
Chore: Config uses env file

saulRM10:
### Description

- `index.js`  will now rely on the updated `config.js ` . 
- updated from .json to .js to leverage the env variables
-  needed the `.sequelizerc` file as outlined in the [sequalize docs ](https://sequelize.org/docs/v6/other-topics/migrations/#the-sequelizerc-file). Essentially the config/ folder and file is needed and created when using `sequalize-cli`, however to make it work with env values we need to use this file set up in the root directory (or wherever the env file exists) 

**NOTE:** 

1. devs need to update their .env file , add the section i added on the env.example but make sure to pass in your credentials (ie db username , password etc) 
2. Now migrations will have to be ran from the root directory (or wherever you have your env file located) !

### Checklist


abanuelo (commented):
LGTM

PR #69, opened 2024-06-04
Feature/division endpoint for league modal

saulRM10:
### Description

- The goal of this endpoint is to find all divisions that are not associated with any session. Meaning they can now be associated with a league
- if non found then it returns an empty array

### Checklist


saulRM10:
no need for this endpoint as we don't save this relationship in the backed without a `season_id`

PR #73, opened 2024-06-06
feat: add utility function to parse raw responses object

garcialuis:
### Description

Adds a utility function to parse raw responses object into individual responses.
The parsing logic also includes functionality to identify duplicates (responses that have already been parsed in the past).

This utility function will later be used by the endpoint that handles retrieving individual responses and serving it to the client.

You can run the utility function by running `node server/utilityFunctions/responseParser.js`


JesusGo115 (approved):
Looks good! Approved!

PR #76, opened 2024-06-09
Feat: Added Form Builder Responses

elimanzo:
### Demo

cb35cd7d05ae27cffdcab3b75d9e50a4

5a7191c770880944e0be33bc01c7eddc

### Description

- Added a form response tab to see the responses 
- Fixed issues with saving form (saving form doesn't recreate fields)

### Issue Link

[Jira Link](https://cascarita.atlassian.net/browse/C1-99)

### Checklist

### Additional Notes

- One thing to note, is that when a form is created and responses start generating, deleting a field will also delete the responses attached to them too (That's just how typeform APIs does it).


client/components/FormResponses/FormResponses.tsx
```diff
@@ -0,0 +1,113 @@
...
+const FormResponses = ({ formId }: FormResponsesProps) => {
+  const [formFields, setFormFields] = useState<Field[]>([]);
+  const [formResponsesMap, setFormResponsesMap] = useState<AnswerRecordMap>(
+    new Map(),
+  );
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const fetchSurveyData = async (endpoint: string) => {
```
abanuelo, line 20: Can we move define this function within the `client/api` folder and import the function into the component?

client/components/FormResponses/FormResponses.tsx
```diff
@@ -0,0 +1,113 @@
...
+                      formResponsesMap.get(responseId)?.get(field.id) as Answer,
+                    )}
+                </td>
+              ))}
+            </tr>
+          ))}
+        </tbody>
+      </table>
+      {formResponsesMap.size === 0 && (
+        <div style={{ display: "flex", justifyContent: "center", padding: 10 }}>
```
abanuelo, line 105: nit: can we add this styling within the `.module.css` file.

client/components/FormResponses/FormResponses.tsx
```diff
@@ -0,0 +1,113 @@
...
+      file_url: () =>
+        answer.file_url ? <a href={answer.file_url}>File</a> : "",
+      boolean: () => (answer.boolean ? "Yes" : "No"),
+      default: () => "",
+    };
+
+    return (typeFormatters[answer.type] ?? typeFormatters.default)();
+  };
+
+  const trimText = (text: string, maxLength: number) => {
```
abanuelo, line 73: Very useful for some of our future frontend components. Can we think about storing this function elsewhere and importing it into the component? I can see if useful for some of the table views we will have!

client/components/FormResponses/FormResponses.tsx
```diff
@@ -0,0 +1,113 @@
...
+      text: () => answer.text ?? "",
+      number: () => String(answer.number ?? ""),
+      email: () => answer.email ?? "",
+      phone_number: () => answer.phone_number ?? "",
+      choice: () => answer.choice?.label ?? "",
+      choices: () => answer.choices?.labels.join(", ") ?? "",
+      date: () =>
+        answer.date ? new Date(answer.date).toLocaleDateString() : "",
+      file_url: () =>
+        answer.file_url ? <a href={answer.file_url}>File</a> : "",
```
abanuelo, line 65: nit: File here can it also be bilingual?

client/components/FormResponses/types.ts
```diff
@@ -0,0 +1,69 @@
...
+    id: string;
+    type: string;
+    ref: string;
+  };
+  type: string;
+  number?: number;
+  text?: string;
+  phone_number?: string;
+  email?: string;
+  date?: string;
```
abanuelo, line 16: Can we stay consistent with backend to frontend types. Like here for instance date is a string when it really should be a timestamp type!

client/components/FormResponses/types.ts
```diff
@@ -0,0 +1,69 @@
...
+  landed_at: string;
+  submitted_at: string;
+  metadata: {
+    user_agent: string;
+    platform: string;
+    referer: string;
+    network_id: string;
+    browser: string;
+  };
+  hidden: Record<string, unknown>;
```
abanuelo, line 38: nit: why do we need an `unknown` type here?

client/pages/Forms/Forms.tsx
```diff
@@ -105,13 +105,16 @@ const Forms = () => {
                   onEdit(
                     form.id,
                     form.title,
-                    form.description,
-                    form.link,
+                    form.welcome_screens[0].properties.description,
```
abanuelo, line 108: Since we are indexing here, is there any risk of `form.welcome_screens` being an empty array? If so this can break! Maybe add some checks for this!

client/pages/NewForm/NewForm.tsx
```diff
@@ -111,14 +113,11 @@ const NewForm = () => {
       const link = surveyResponseObj._links.display;
       setSurveyLink(link);
       setFormId(surveyId);
-      existingSurveys[formId ?? surveyId] = {
-        id: formId ?? surveyId,
+      setFields(surveyResponseObj.fields);
+      existingSurveys[surveyId] = {
+        ...surveyResponseObj,
         edittedBy: currentUser?.first_name ?? "",
```
abanuelo, line 119: nit: should this be `editedBy` instead of `edittedBy`?

client/pages/NewForm/NewForm.tsx
```diff
@@ -156,16 +155,14 @@ const NewForm = () => {
...
       surveys[formId] = {
-        ...surveys[formId],
+        ...surveyResponseObj,
         edittedBy: currentUser?.first_name ?? "",
         lastUpdated: new Date().toLocaleString(),
-        title,
-        description,
-        ...data,
       };
       localStorage.setItem("surveys", JSON.stringify(surveys));
```
abanuelo, line 167: nit: Maybe we can make a new issue on this, but I see that we have some reliance on localStorage variables and if this continues to grow, we need to locate at different places in our code to keep track what is being set where. Let's consider making a `localStorage.ts` file where we keep all references to setting localStorage variables and values, etc. Also once we logOut maybe we should consider unsetting them within that same file!

client/pages/NewForm/NewForm.tsx
```diff
@@ -176,13 +173,7 @@ const NewForm = () => {
...
     <Page>
       <div>
-        <div
-          className={styles.newFormHeader}
-          style={{
-            borderBottom: "1px solid #DFE5EE",
-            marginBottom: 15,
-            marginRight: 33,
-          }}>
+        <div className={styles.newFormHeader} style={{}}>
```
abanuelo, line 176: nit: remove `style`

client/pages/NewForm/NewForm.tsx
```diff
@@ -208,58 +199,82 @@ const NewForm = () => {
...
-          </div>
-          <div className={styles.formCanvasContainer}>
-            <div className={styles.formTitleContainer}>
-              <div style={{ paddingBottom: 8 }}>
+        <ul className={styles.formNav}>
+          <li
+            className={
+              activeSection === "questions" ? styles.activeSection : ""
+            }
+            style={{ marginRight: 20 }}
```
abanuelo, line 207: nit: add that style within the className

client/pages/NewForm/NewForm.tsx
```diff
@@ -208,58 +199,82 @@ const NewForm = () => {
...
+              {t("formNavOptions.responses")}
+            </li>
+          )}
+        </ul>
+        {activeSection === "questions" && (
+          <div className={styles.newFormContainer}>
+            <div className={styles.formElementsContainer}>
+              <h2 className={styles.subtitle}>{t("formElements")}</h2>
+              <hr />
+              <p className={styles.smallText} style={{ paddingTop: 8 }}>
```
abanuelo, line 226: nit: add style to className

abanuelo (changes requested):
Overall great work! Some minor nits and future work items as we continue to scale the frontend of our app! Thanks for making these needed changes!

client/pages/NewForm/NewForm.tsx
elimanzo: Local storage is a temp fix for storing the forms created by user. The is currently being worked on by the backend, to be able to store and fetch forms created by users / groups.

I will create an issue for now. 👍🏽

client/components/FormResponses/types.ts
elimanzo: Hidden fields are custom data you can pass to Typeform responses, such as user IDs, session IDs, or other metadata that you want to keep track of. Since these fields can depend on the specific form and use case I resorted to using `unknown`.

abanuelo: Got you @elimanzo this makes sense to me!

abanuelo (changes requested):
Great work on addressing all the comments! Approving on my end!

client/pages/NewForm/NewForm.tsx
abanuelo: Issue has been created resolving this conversation!

elimanzo:
Removed local storage calls for #78

abanuelo (approved):
LGTM!

PR #77, opened 2024-06-10
Feature/league dashboard

jipatino98:
### Description

This PR introduces the League Dashboard as well as a reusable component, `DashboardTable`, that will allow developers to create a consistent and styled table to display data in their dashboard. Documentation for this component will be updated in the [Confluence Component Documentation](https://cascarita.atlassian.net/wiki/spaces/~71202069510b02626642319e93c473c7e23d26/pages/32309250/Component+Documentation) file.

This PR also integrated [TanStack Query's](https://tanstack.com/query/latest/docs/framework/react/overview) fetching library that will allow us to, in addition to fetching data from our apis, cache our queries and automatically refresh data when it becomes stale.

### Backend Changes
**In league.controller.js:**
- Originally, if a league_id returns an object with no leagues, the request returns an 505 server error rather than returning a response with no leagues. See the snapshot below to see the changes to the file inside the `if` block.

 
Screenshot 2024-06-10 at 2 20 45 PM

**In league.route.test.js and group.route.test.js**
- I made some changes to the test suites. Instead of returning an error when there are no leagues for a group_id, it instead expects an empty array.

### Issue Link
- **Create League Dashboard Ticket**
https://cascarita.atlassian.net/browse/C1-87

- **Link "Add League" Button to Leagues Dashboard Ticket**
https://cascarita.atlassian.net/browse/C1-85

### Checklist


client/api/teams/types.ts
```diff
@@ -6,4 +6,29 @@ interface TeamResponse {
...
+  //Use to set open state from true to false after form submission
+  afterSave: () => void;
+}
+
+interface LeagueType {
+  id: number;
+  group_id: number;
+  name: string;
+  description: string;
+  created_at: string;
```
abanuelo, line 19: I know that on the backend these are official timestamp types, can we preserve that typing on the frontend as well?

client/components/Forms/LeagueForm.tsx
```diff
@@ -1,42 +1,58 @@
...
+import { User } from "../AuthContext/types";
 
 interface LeagueFormProps {
   //Use to set open state from true to false after form submission
   afterSave: () => void;
+  currentUser: User;
 }
 
-const LeagueForm: React.FC<LeagueFormProps> = ({ afterSave }) => {
+const addLeague = async (formData: object) => {
```
abanuelo, line 13: Can we add this inside the `client/api` folder?

saulRM10, line 13: move api call

client/components/Forms/LeagueForm.tsx
```diff
@@ -1,42 +1,58 @@
...
+  const queryClient = useQueryClient();
+
+  const leagueFormMutation = useMutation({
+    mutationFn: addLeague,
+    onSuccess: () => {
+      queryClient.invalidateQueries({
+        queryKey: ["leagues"],
+      });
+    },
+  });
```
abanuelo, line 38: This is just a general question for my understanding. Here when we call `useMutation()`, are we storing the result in a key value dictionary in local memory with the key being `leagues` and the response being the `value`?

client/pages/Leagues/Leagues.tsx
```diff
@@ -4,35 +4,43 @@ import PrimaryButton from "../../components/PrimaryButton/PrimaryButton";
...
+import LeagueForm from "../../components/Forms/LeagueForm";
+import { LeagueType } from "../../api/teams/types";
+import { useQuery } from "@tanstack/react-query";
+import DashboardTable from "../../components/DashboardTable/DashboardTable";
 import { useState } from "react";
 import { useTranslation } from "react-i18next";
+import { User } from "../../components/AuthContext/types";
 
-const Leagues = () => {
+interface LeaguesProps {
```
abanuelo, line 16: Can we add this Interface into its own `types.ts` file within the directory as the `Leagues.tsx` folder?

client/pages/Leagues/Leagues.tsx
```diff
@@ -4,35 +4,43 @@ import PrimaryButton from "../../components/PrimaryButton/PrimaryButton";
...
-    "test",
-    "test",
-    "test",
-  ];
-
   const filterStatuses = [t("filterOptions.item1"), t("filterOptions.item2")];
   const sortStatuses = [t("sortOptions.item1"), t("sortOptions.item2")];
+  const [open, setOpen] = useState(false);
+
+  const leaguesQuery = useQuery({
```
abanuelo, line 30: For these queries, we also consider storing them within the `client/api` folder.

client/pages/Leagues/Leagues.tsx
```diff
@@ -4,35 +4,43 @@ import PrimaryButton from "../../components/PrimaryButton/PrimaryButton";
...
+    queryFn: () =>
+      fetch(`/api/groups/${currentUser.group_id}/leagues`).then((res) =>
+        res.json(),
+      ),
+    queryKey: ["leagues"],
+  });
+
+  console.log(currentUser);
```
abanuelo, line 38: nit: remove `console.log`

client/pages/Leagues/Leagues.tsx
```diff
@@ -69,22 +78,46 @@ const Leagues = () => {
...
+            <LeagueForm
+              afterSave={() => setOpen(false)}
+              currentUser={currentUser}
+            />
+          </Modal.Content>
+        </Modal>
       </div>
+
+      {leaguesQuery.data == null || leaguesQuery?.data?.length === 0 ? (
+        <p className={styles.noLeagueMessage}>Add a League to Display...</p>
```
abanuelo, line 98: nit: maybe we can push this to another PR, but since our app is now supporting multiple languages, can we make this text follow the convention of other Frontend files to generate both Spanish + English text?

abanuelo (changes requested):
Overall nice work! I learned about query caching! Thanks for making these changes.

client/components/Forms/LeagueForm.tsx
jipatino98: Yeah so `useMutiation` takes in a few option parameters. The ones i used above:
- `mutationFn` is the asynchronous function that returns the promise
- `onSuccess` is one of the various option that tells the React Query that once the mutation is successful (ie the request was successful and the data has been modified), any query with the key `leagues` is refetched and revalidated under the hood.
React Query's hook like `useQuery` and `useMutation` manage all the caching, storing, and refreshing in the background under the hood.

client/pages/Leagues/Leagues.tsx
jipatino98: Yes! In the meantime, I can commit what i have. After, we can meet and show me how to use the translation function!

client/api/leagues/service.tsx
```diff
@@ -0,0 +1,42 @@
...
+  try {
+    const response = await fetch(`/api/groups/${groupId}/leagues`, {
+      method: "GET",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      mode: "cors",
+    });
+    const result = response.json();
+    return result;
```
elimanzo, line 18: you could just return `response.json()`

client/components/Forms/LeagueForm/LeagueForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+      new FormData(event.currentTarget),
+    );
+
+    const newLeague = {
+      name: leagueName,
+      description: leagueDescription,
+      group_id: currentUser.group_id,
+    };
+
+    leagueFormMutation.mutate({
```
elimanzo, line 35: i think we can delete `newLeague`

and just do 


```
leagueFormMutation.mutate({
      name: leagueName,
      description: leagueDescription,
      group_id: currentUser.group_id,
});
```

client/pages/Home/Home.tsx
```diff
@@ -4,7 +4,9 @@ import { useAuth } from "../../components/AuthContext/AuthContext";
 const Home = () => {
   const { currentUser } = useAuth();
 
-  return <div>{currentUser ? <Leagues /> : <></>}</div>;
+  return (
+    <div>{currentUser ? <Leagues currentUser={currentUser} /> : <></>}</div>
```
elimanzo, line 8: is we're using a hook for `currentUser` why do we need to prop drill here?

server/controllers/league.controller.js
```diff
@@ -1,20 +1,22 @@
...
+const modelByPk = require("./utility");
 
 const LeagueController = function () {
   var getLeagueByGroupId = async function (req, res, next) {
     const groupId = req.params.id;
 
     try {
-      const result = await League.findAll({
+      await modelByPk(res, Group, groupId);
+      var result = await League.findAll({
```
elimanzo, line 12: `var` makes me assume that we are going to be changing this value outside of the scope. I think changing it to `let` would be suitable here.

elimanzo (changes requested):
Good job using React Query! I had some nit feedback and some questions as well

client/api/leagues/service.tsx
jipatino98: Good catch! Resolve that!

client/pages/Home/Home.tsx
jipatino98: You are correct, we dont need to! I will change so that we are calling the hook rather than prop drilling.

client/api/teams/types.ts
jipatino98: In JavaScript, there isn't a specific timestamp type like in SQL databases, but I can use the `Date` object to handle date and time values. It would look some thing like:
``` javascript
interface Group {
  group_id: number;
  name: string;
  description: string;
  created_at: Date;
} 
```
Then once we fetch the data from the backend, I would just have to make sure to parse the created_at field into a Date object. I will commit the changes, ensuring that `created_at` is a Date object instead.

server/controllers/league.controller.js
```diff
@@ -1,20 +1,22 @@
...
+const modelByPk = require("./utility");
 
 const LeagueController = function () {
   var getLeagueByGroupId = async function (req, res, next) {
     const groupId = req.params.id;
 
     try {
-      const result = await League.findAll({
+      await modelByPk(res, Group, groupId);
+      let result = await League.findAll({
```
elimanzo, line 12: Looking back at this again, we can just make this a `const`

and just 
`return res.status(200).json(Object.keys(result).length === 0 ? [] : result);`

elimanzo (commented):
Looks good on my end, good job!

server/controllers/league.controller.js
jipatino98: That actually looks a lot cleaner. I can't remember the reason we used `let` or `var` in the first place, but assuming we are not changing the value anywhere else in the file, `const` does sound like the better declaration. I'll make that change right now. Thank you Eli!

client/api/leagues/service.tsx
```diff
@@ -0,0 +1,41 @@
...
+    return response.json();
+  } catch (error) {
+    console.error("Error fetching league: ", error);
+    throw error;
+  }
+};
+
+const createNewLeague = async (formData: object) => {
+  try {
+    const response = await fetch("/api/leagues/", {
```
abanuelo, line 26: nit: remove trailing `/`.

abanuelo (approved):
Just one nit, but after feel free to merge this pull request looks great!

elimanzo (approved):
Looks Good!

PR #80, opened 2024-06-12
Feature: Add GET Form endpoints 

saulRM10:
### Description

- route gets all the forms for a from a group id.
- update the `Forms` sql table to include user id for whoever created the form and whoever updated the form 
- need to delete all your SQL Forms data and run migrations 
- need to delete all mongo db clusters so its all in sync
- one thing to note is that we only looking up existing forms from our SQL `Forms` table , not making a GET request to typeform 

### Checklist

### DIFF
--- a/server/controllers/form.controller.js
+++ b/server/controllers/form.controller.js
@@ -8,9 +8,33 @@ const RawResponse = require("./../mongoModels/raw_response");
 const ResponseId = require("./../mongoModels/response_id");
 const Response = require("./../mongoModels/response");
 const FormMongo = require("./../mongoModels/form");
-const { Form } = require("../models");
+const { Form, User } = require("../models");
 
 const FormController = {
+  async getAllForms(req, res, next) {
+    try {
+      const groupId = req.params.id;
+
+      const sqlForms = await Form.findAll({
+        where: {
+          group_id: groupId,
+        },
+        attributes: ["document_id"],
+      });
+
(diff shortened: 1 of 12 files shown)


server/controllers/form.controller.js
```diff
@@ -104,17 +117,37 @@ const FormController = {
...
+        group_id: req.params.group_id,
+        created_by: createdBy,
+        updated_by: null,
         form_data: responseBody,
         form_blueprint: req.body,
       });
 
       const result = await form.save();
 
+      console.log("Testing the group-id thingy", req.params.group_id);
```
garcialuis, line 145: Remove this

server/migrations/20240611200740-update-form-sql-table-add-user-columns.js
```diff
@@ -0,0 +1,49 @@
...
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    const transaction = await queryInterface.sequelize.transaction();
+
+    try {
+      await queryInterface.addColumn("Forms", "created_by", {
+        type: Sequelize.INTEGER,
+        allowNull: true,
```
garcialuis, line 11: I think `created_by` should not be null, we know who created the form.

garcialuis (changes requested):

garcialuis (approved):
LGTM!

PR #81, opened 2024-06-13
Feature/update forms endpoint

saulRM10:
### Checklist

### DIFF
--- a/server/controllers/form.controller.js
+++ b/server/controllers/form.controller.js
@@ -11,6 +11,61 @@ const FormMongo = require("./../mongoModels/form");
 const { Form, User } = require("../models");
 
 const FormController = {
+  async updateForm(req, res, next) {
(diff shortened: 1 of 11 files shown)


server/controllers/form.controller.js
```diff
@@ -8,9 +8,88 @@ const RawResponse = require("./../mongoModels/raw_response");
...
+      },
+    );
+
+    return res.status(201).json(updatedForm);
+  },
+  async getAllForms(req, res, next) {
+    try {
+      const groupId = req.params.id;
+
+      const sqlForms = await Form.findAll({
```
JesusGo115, line 73: You should query directly from the mongo DB instead of using the SQL database. Mongo also has group_id in the forms

JesusGo115 (changes requested):

PR #85, opened 2024-06-20
Fix: fixes duplicate individual form responses

garcialuis:
### Description
This PR fixes the issue where individual form responses were being duplicated.

Fix:
- Adds check to ensure that if we have no new responses we do not create an empty responseids document.

### Checklist


JesusGo115 (approved):
Let's good!

PR #90, opened 2024-06-27
fix: adjust z-index for top nav

abanuelo:
### Description

TopNavBar would still show even when modal popped up. Removing z-index solves the problem.

https://github.com/cascaritaco/cascarita/assets/32311654/76855d0c-64e4-4d19-902b-de902c635b4a

### Checklist

- [X] I have tested the changes locally by running `npm run test`
- [ ] I have added or updated relevant documentation
- [ ] I have autoformatted the code by running `npm run eslint`
- [ ] I have added test cases (if applicable)

PR #92, opened 2024-07-04
fix: correct z-index to show language options

abanuelo:
### Description

Although the z-index bug was fixed in #90 , it conflated with showing the `LanguageDropDown` component so that it would not show. That has been corrected. Also timeout was added to the `LanguageDropDown` to stop showing if no activity to select a language was made over the last 5 seconds.

https://github.com/cascaritaco/cascarita/assets/32311654/f74bce89-9a9d-4aac-a5ee-60ca541cfa54

### Checklist

### DIFF
--- a/client/components/Modal/Modal.module.css
+++ b/client/components/Modal/Modal.module.css
@@ -46,6 +46,7 @@
 
 .overlay[data-state="open"] {
   animation: dialog-overlay-show 200ms;
+  z-index: 2;
 }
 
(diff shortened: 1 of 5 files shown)


client/components/Modal/Modal.module.css
jipatino98, line 1: Good call on adding a z-index here. Ensures that modal will always be on top layer.

jipatino98 (approved):
Looks good to me! I will approve this change, good work

PR #93, opened 2024-07-05
Feat: Create Response Endpoint

JesusGo115:
### Description

Added the createResponse endpoint to insert responses into the MongoDB and updated the getResponses endpoint to retrieve all the responses from the MongoDB

### Checklist


server/utilityFunctions/responseParser.js
```diff
@@ -1,4 +1,6 @@
-const ResponseId = require("../mongoModels/response_id");
+// ** DEPRECATED **
```
abanuelo, line 1: nit: but why is this deprecated, I am fine leaving the comment here but giving some more context as to why that is would be helpful!

abanuelo (changes requested):
Overall good stuff! Just a minor comment. Also maybe make an issue for this for future reference but `/:form_id/responses` should ideally be paginated so that we don't return too many responses all at once from a league. For now, this looks good!

server/controllers/form.controller.js
```diff
@@ -4,8 +4,6 @@ require("dotenv").config();
 
 const jsonParser = require("../utilityFunctions/responseParser");
```
saulRM10, line 5: remove this if you not using it no more

saulRM10 (approved):
Over all looks good

server/utilityFunctions/responseParser.js
saulRM10: if depreciated,  should we just get rid of the file @garcialuis ?

JesusGo115: Hey Armando, this file is deprecated since it was primarily used for the endpoints that dealt with Typeform. Since we're no longer using Typeform, this file and its functions aren't needed at the moment. I wasn't too sure if I should delete it, it might deem useful later, but we'll wait for @garcialuis and see what he thinks.

server/controllers/form.controller.js
JesusGo115: Removed!

client/pages/FormPage/FormPage.tsx
```diff
@@ -53,7 +55,12 @@ const FormPage = () => {
         };
       }) ?? [];
     // TODO: send this to the backend as an API call
-    return normalizedAnswers;
+    console.log(normalizedAnswers);
```
elimanzo, line 58: lets remove console log.

client/pages/FormPage/FormPage.tsx
```diff
@@ -53,7 +55,12 @@ const FormPage = () => {
...
+    const responsesData = await createMongoResponse(
+      "testFormId",
```
elimanzo, line 60: don't need to pass this string, form's page already takes in the formId, so you can just pass in `formId`

elimanzo (changes requested):
looks good just couple of nit 😃

client/pages/FormPage/FormPage.tsx
JesusGo115: Changed!

client/pages/FormPage/FormPage.tsx
JesusGo115: Removed!

server/utilityFunctions/responseParser.js
JesusGo115: Deleted the file!

elimanzo (approved):
LGTM Good job!

server/controllers/form.controller.js
```diff
@@ -24,77 +20,29 @@ const FormController = {
...
-      const responsesToInsert = sortedResponses.response.map((res) => ({
-        form_id: req.params.form_id,
-        response: res,
-      }));
-
-      await Response.insertMany(responsesToInsert);
-
-      const individualResponses = await Response.find({
+      return res.status(201).json(insertedResponse);
+    } catch (error) {
```
abanuelo, line 35: should we have a `console.error()` here?

abanuelo (approved):

client/pages/FormPage/FormPage.tsx
```diff
@@ -52,8 +53,12 @@ const FormPage = () => {
           type: answerType,
         };
       }) ?? [];
-    // TODO: send this to the backend as an API call
-    return normalizedAnswers;
+
+    const responsesData = await createMongoResponse(
```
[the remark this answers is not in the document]
JesusGo115: Added a try/catch to this

server/controllers/form.controller.js
JesusGo115: The middleware has the console.error() and the next(error) triggers the middleware.

PR #95, opened 2024-07-07
Poc/set up stripe 

saulRM10:
### Description

- set up the backend for stripe integration; there are 3 crucial phases that we need to build out and I build out 2 of those 3 but just the backend part.  Backend logic can be found in `account.routes.js `and `account.controller.js`
- League Admin = whoever creating the form and wanting to collect payments

**_Phase 1: League Admin links/creates stripe account flow:_**

1. This phase 1 is  to redirect the League Admin to Stripe Portal to link or create a stripe account ( i have a front end component for that but we need to redesign and workout that ui still, `StripeConnectButton.tsx`) 
2.  Relevant backend code is the `AccountController.createAccountConnection()` does a couple of things. It creates a stripe account for the user using `Stripe.accounts.create()` , which we for now (at that point) only care for the generated id `act_123ABC`. We use that id to generate a temp link that is returned to the front end to redirect the user to the stripe dashboard and then the user either links a stripe account or create a new account. I created a migration to create the `UserStripeAccounts` table, which stores the League Admin id (id we use in cascarita) and their stripe account id (`act_123ABC` ). So we can create that relationship between a cascarita user and their stripe account. 

**_Phase 2: League Admin creates a 'product' :_**

1. Backend logic only: `AccountController.createPaymentIntent()`, Using [Stripe Payment Intents ](https://docs.stripe.com/api/payment_intents)
2.  create a payment intent, store that intent along with the form id , user id all in a new table i wrote a migration for `FormPaymentIntents` to keep the track of the relationships of payments intents by form and user.

**_Phase 3: user pays using the form League Admin created flow :_**

1. User uses our custom component that uses stripe.js payment elements to make a payment on the form
2. This flow requires a lot of front end, since stripe has payment elements. This phase  is yet to be worked on both front end and backed. 

> Collect payment details on the client with the [Payment Element](https://docs.stripe.com/payments/payment-element). The Payment Element is a prebuilt UI component that simplifies collecting payment details for a variety of payment methods.
The Payment Element contains an iframe that securely sends payment information to Stripe over an HTTPS connection.


client/pages/Forms/Forms.tsx
```diff
@@ -107,6 +108,7 @@ const Forms = () => {
           </div>
         </div>
         <PrimaryButton label={t("button")} onClick={handleNewFormClick} />
+        <ConnectWithStripeButton />
```
JesusGo115, line 111: I'm not 100% sure if this is the cause, but due to the changes, it doesn't allow you to open the form to submit a response

server/controllers/account.controller.js
```diff
@@ -0,0 +1,106 @@
...
+  };
+
+  var getClientSecret = async function (req, res, next) {
+    try {
+      const paymentIntentId = req.params["paymentIntentId"];
+      const stripeAccountId = await getStripeAccountId(
+        req.params["account_id"],
+      );
+
+      console.log(stripeAccountId);
```
JesusGo115, line 83: remove the console.log

server/models/form.js
```diff
@@ -21,6 +21,10 @@ module.exports = (sequelize, DataTypes) => {
         foreignKey: "updated_by",
         targetKey: "id",
       });
+      Form.hasMany(models.FormPaymentIntents, {
+        foreignKey: "form_id",
+        targetKey: "id",
```
JesusGo115, line 26: Should be sourcekey instead of targetkey

JesusGo115 (changes requested):
Overall, looks good, but there are some issues that need to be fixed

client/pages/Forms/Forms.tsx
saulRM10: It's not due to the stripe button component

JesusGo115 (approved):
Looks good!

PR #98, opened 2024-07-22
fix: docker deployments to ecs

abanuelo:
### Description

Looking to change some of the format for Docker files present in our repo to facilitate deployment!

### Checklist


docker-compose.yml
```diff
@@ -2,33 +2,34 @@ version: "3.8"
...
     build:
       context: .
       dockerfile: server/Dockerfile
+    volumes:
+      - ./server:/app/server
+      - ./dist:/app/dist
     # NOTE, if you change the .env SERVER_PORT, please update here to match
     # Please update in the corresponding Dockerfile as well
     ports:
       - "3001:3001"
```
alexrdrgz, line 17: You can do the same things here

docker-compose.yml
```diff
@@ -2,33 +2,34 @@ version: "3.8"
...
       - "3306:3306"
-    env_file:
-      - .env
-    volumes:
-      - .:/app
```
alexrdrgz, line 16: You may want to decouple the volumes when deploying to production by using named volumes. 

I don't know the full context of the project but if you want the data to persist when containers are restarted or not you'd want to take that path

alexrdrgz (commented):
Small optimizations but looks like a good refactor

Note: did not test the container

client/Dockerfile
```diff
@@ -2,4 +2,12 @@ FROM node:21-alpine
 
 WORKDIR /app
 
-EXPOSE 3000
\ No newline at end of file
+COPY ./package.json ./package-lock.json ./webpack.config.js /.env ./
```
alexrdrgz, line 5: Ultra nit picky but you can optimize this to not include the build tools when deploying to production.

Makes it slightly leaner and more secure

alexrdrgz: meant to comment this on the server

docker-compose.yml
```diff
@@ -2,33 +2,34 @@ version: "3.8"
...
+    image: cascarita-client
     build:
       context: .
       dockerfile: client/Dockerfile
-    env_file:
-      - .env
     # NOTE, if you change the .env CLIENT_PORT, please update here to match
     # Please update in the corresponding Dockerfile as well
     ports:
       - "3000:3000"
```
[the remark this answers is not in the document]
abanuelo: Thanks for this! We definitely want a way to sync with the `env` file values since that is going to be the source of truth!

docker-compose.yml
abanuelo: Wait, can you elaborate by what you mean "same things here"?

client/Dockerfile
abanuelo: Gotcha, let me try to remove that from the server Docker file so that can make it leaner and more secure thanks for the feedback!

server/Dockerfile
```diff
@@ -2,11 +2,19 @@ FROM node:21-alpine
...
+# Needed for the bcrypt dependency
 RUN apk add --no-cache make gcc g++ python3 && \
-    npm install && \
     npm rebuild bcrypt --build-from-source && \
     apk del make gcc g++ python3
 
-EXPOSE 3001
\ No newline at end of file
+EXPOSE 3001
+EXPOSE 3306
```
garcialuis, line 18: We can remove this line, we should not expose port 3306 from the server container

docker-compose.yml
```diff
@@ -2,33 +2,34 @@ version: "3.8"
...
       context: .
       dockerfile: server/Dockerfile
+    volumes:
+      - ./server:/app/server
+      - ./dist:/app/dist
     # NOTE, if you change the .env SERVER_PORT, please update here to match
     # Please update in the corresponding Dockerfile as well
     ports:
       - "3001:3001"
       - "3306:3306"
```
garcialuis, line 18: We should also remove port 3306 from the docker-compose file

garcialuis (changes requested):

docker-compose.yml
abanuelo: Ack! I have resolved the communication update here! Closing out the PR!

server/Dockerfile
abanuelo: Ack! Done!

docker-compose.yml
abanuelo: Ack! Done!

garcialuis (approved):
Thanks for making those changes Armando, LGTM

PR #104, opened 2024-08-06
2FA Auth Code Generation/Verification

JesusGo115:
### Description

Created backend logic to generate a 2FA code sent through mail to verify your account before resetting your password. 

### Testing the Features
**Make sure to do _npm i_ and _run sequelize migrations_**

The route for sending an email and verifying an auth code can be found under "User Auth" on postman. 

You need a **Brevo** account and an api key. In your .env file, add "BREVO_API_KEY=" and your api key here

**Make sure to modify this before testing:**
1. in `user.controller.js,` change the `senderEmail` to your email used for your Brevo account **It will not work otherwise**
2. Within the postman, change `email` within the body tab to the recipient email on both `send email` and `verify auth code` routes

### Things to note

- Email function was moved to the user controller as the function can be used for both sending form links and auth code, depending on the route used (look at the `sendOtpEmail` and `sendFormLinkEmail` functions within the `user.controller.js` file to view email formats)
- New table has been added `AuthCode` to keep track of all auth codes so make sure to run migrations
- Added new update PATCH function in the `user.controller.js` based on email to help with reseting a password. It can viewed in postman

### Checklist

- [x] Backend logic to reset the password once the auth code was verified


server/controllers/user.controller.js
```diff
@@ -115,11 +138,146 @@ const UserController = function () {
     }
   };
 
+  var sendEmail = async function (req, res, next) {
```
saulRM10, line 141: Do the forms use this function as well no ? I see you deleted some email function from the forms controller

abanuelo: I think when I created it, I added it as part of the forms controller but honestly its pretty independent so feel free to include within the users controller file if it makes sense!

server/controllers/user.controller.js
```diff
@@ -115,11 +138,146 @@ const UserController = function () {
...
+    try {
+      const apiKey = process.env.BREVO_API_KEY;
+      const url = "https://api.brevo.com/v3/smtp/email";
+
+      const emailType = req.body.emailType;
+      const typeformLink = req.body.formLink || "";
```
abanuelo, line 147: nit: since we already removed all typeform dependencies, maybe rename this to `formLink`? 
```suggestion
      const formLink = req.body.formLink || "";
```

server/controllers/user.controller.js
```diff
@@ -115,11 +138,146 @@ const UserController = function () {
...
+      }
+
+      if (userOtp.expiration_date < new Date()) {
+        res.status(401);
+        throw new Error("auth code has expired. generate a new code");
+      }
+
+      const result = await compareOtp(otp, userOtp.code);
+
+      if (!result) {
```
abanuelo, line 251: I see we are storing attempts. Are there any best practices for how to limit the userOpt code entry attempts? Like should we stop it after `x` attempts? If so consider the following suggestion. 
```suggestion
      const MAX_OPT_ATTEMPTS = 5;
      if (!result || userOtp.attempts >= MAX_OPT_ATTEMPTS) {
```

abanuelo (changes requested):
Honestly, great job @JesusGo115! Code is clean, readible, and the logic is sound! Just a couple of nits but once those have been resolved happy to approve!

server/controllers/user.controller.js
```diff
@@ -115,11 +138,146 @@ const UserController = function () {
...
+      const emailType = req.body.emailType;
+      const typeformLink = req.body.formLink || "";
+      const recipientEmail = req.body.email;
+
+      const senderEmail = "jgomez.cascarita@gmail.com"; //Ask Armando if we could use a fake email instead of ours (i.e. no-reply-cascarita@gmail.com)
+      const senderName = "Cascarita";
+      const type = "classic";
+
+      const authCode =
+        emailType === "passwordCode" ? await generateOTP() : null;
```
saulRM10, line 155: this generateOTP should not be asynchronous, its a small operation.

saulRM10 (changes requested):
Left some comments. The generation of emails is using a lot of hardcoded values. What id we add more types of emails ? Also the `sendEmail` is doing a lot , maybe break it up ; `storeCode()` etc..

server/controllers/user.controller.js
JesusGo115: It was originally in the forms controller, like Armando mentioned, but since we plan to send emails with different formats (One for sending form links and another for auth codes), I thought it'd be more appropriate to move it to User controller as it has uses for both forms and user information.

saulRM10: Makes sense, my concern was if you removed it from the form controller. We can still send form links via email

server/controllers/user.controller.js
saulRM10: agree

server/controllers/user.controller.js
JesusGo115: You right. Changed.

server/controllers/user.controller.js
JesusGo115: Changed.

server/controllers/user.controller.js
JesusGo115: Added!

server/controllers/user.controller.js
```diff
@@ -138,84 +138,121 @@ const UserController = function () {
...
+
+      req.params = { id: user.id };
+
+      const response = await updateUserById(req, res, next);
+    } catch (error) {
+      next(error);
+    }
+  };
+
+  var sendEmail = async function (recipientEmail, subject, emailContent) {
```
saulRM10, line 164: I think we should have this in a try/catch and in the catch throw the error. I see your if statement checking for `response.status` but i think we should know exactly when this third party api fails. Also i see you pass back the responseData but don't use it, another reason to just have a try/catch here in `sendEmail()` if an error occurs it will throw it and whoever called it will know without needed to have an if statement. If email is sent successfully return nothing.

server/controllers/user.controller.js
```diff
@@ -138,84 +138,121 @@ const UserController = function () {
...
+          <html>
+            <body>
+              <p>Hello,</p>
+              <p>Use this code to reset your password:</p>
+              <h1><strong>${authCode}</strong></h1>
+              <p>Thank you!</p>
+            </body>
+          </html>`;
+
+      const response = await sendEmail(recipientEmail, subject, emailContent);
```
saulRM10, line 211: if you add a try/catch in the` sendEmail() ` all you have to do is await that sendEmail to finish up, if no error then function continues to execute. If error with `sendEmail()` your try/catch in **this** function will get hit. Thus eliminating the need to return and store a response from `sendEmail()` and also eliminate the need to do an `if` check one the status

server/controllers/user.controller.js
```diff
@@ -138,84 +138,121 @@ const UserController = function () {
...
+          code: await hashOtp(authCode),
+          attempts: 0,
+          start_date: new Date(),
+          expiration_date: new Date(new Date().getTime() + 15 * 60000),
+        };
+
+        await AuthCode.build(newAuthCode).validate();
+        await AuthCode.create(newAuthCode);
+      } else {
+        throw new Error(
```
saulRM10, line 225: throw error in the `sendEmail()`

server/controllers/user.controller.js
```diff
@@ -138,84 +138,121 @@ const UserController = function () {
...
-      const response = await fetch(url, {
-        method: "POST",
-        headers: {
-          Accept: "application/json",
-          "Content-Type": "application/json",
-          "api-key": apiKey,
-        },
-        body,
-      });
+      const response = await sendEmail(recipientEmail, subject, emailContent);
```
saulRM10, line 253: if you update `sendEmail() `to throw error, no need to store the whole thing in a `response`

server/controllers/user.controller.js
```diff
@@ -115,11 +138,191 @@ const UserController = function () {
...
+      if (!user) {
+        res.status(404);
+        throw new Error(`no user was found with email ${user.email}`);
+      }
+
+      delete req.body.email;
+
+      req.params = { id: user.id };
+
+      const response = await updateUserById(req, res, next);
```
saulRM10, line 158: A bit confused by this one, you find the user by email with a `User.findOne()` and then call updateUserById where that one does another query with `User.findByPk()`

saulRM10 (changes requested):
I like that you broke it down, just requesting changes on the way you handle the error in the `sendEmail()` , also left a comment on a function i am confused by: `updateUserByEmail()`

server/controllers/user.controller.js
```diff
@@ -115,11 +138,191 @@ const UserController = function () {
...
+    } catch (error) {
+      next(error);
+    }
+  };
+
+  var sendEmail = async function (recipientEmail, subject, emailContent) {
+    const apiKey = process.env.BREVO_API_KEY;
```
abanuelo, line 165: nit: can we be specific as to why API Key this is, `brevoAPIKey`?
```suggestion
    const brevoAPIKey = process.env.BREVO_API_KEY;
```

server/routes/user.routes.js
```diff
@@ -9,14 +9,18 @@ router.get("/loginReactPageHere", (req, res) => {
...
   "/login",
   passport.authenticate("local", {
     failureRedirect: "/user/loginReactPageHere",
   }),
-  UserController.logInUser
+  UserController.logInUser,
 );
+router.post("/otpEmail", UserController.sendOtpEmail);
+router.post("/formEmail", UserController.sendFormLinkEmail);
+router.post("/verifyOTP", UserController.verifyOTP);
```
abanuelo, line 24: These don't seem restful. Can we change them to as follows:

```suggestion
router.post("/otp/emails", UserController.sendOtpEmail);
router.post("/forms/emails", UserController.sendFormLinkEmail);
router.post("/otp/verification", UserController.verifyOTP);
```

abanuelo (changes requested):
Added some more review here as well! Great work! It is coming together team!

server/routes/user.routes.js
JesusGo115: Changed! Thanks for the heads up!

server/controllers/user.controller.js
JesusGo115: Changed!

server/controllers/user.controller.js
JesusGo115: Met up with Saul, worked out a solution!

server/controllers/user.controller.js
JesusGo115: Removed the updateUserByEmail as it's redundant.

saulRM10 (approved):
Looks good, great job Chuy !

abanuelo (approved):
LGTM! Thanks for being thorough with the feedback!

PR #111, opened 2024-08-13
feat: adding working search and filtering

abanuelo:
### Description

Now that table views for most pages have been developed, working to add working filter and search bars! Also removing unnecessary routes and components to keep UI clean:

https://github.com/user-attachments/assets/6e8ec6c3-e383-4131-ba9e-ae110cb46d3d

### Checklist

### DIFF
--- a/client/pages/Forms/Forms.tsx
+++ b/client/pages/Forms/Forms.tsx
@@ -52,6 +52,7 @@ const Forms = () => {
   const navigate = useNavigate();
   const { currentUser } = useAuth();
   const sortStatuses = [t("sortOptions.item1"), t("sortOptions.item2")];
+  const [searchQuery, setSearchQuery] = useState("");
 
   useEffect(() => {
     (async () => {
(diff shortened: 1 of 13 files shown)


client/pages/Forms/Forms.tsx
```diff
@@ -90,12 +91,27 @@ const Forms = () => {
...
+      if (sorts === t("sortOptions.item1")) {
+        return a.form_data.title.localeCompare(b.form_data.title);
+      } else if (sorts === t("sortOptions.item2")) {
+        return (
+          new Date(b.updatedAt).getTime() - new Date(a.updatedAt).getTime()
+        );
+      }
+      return 0;
+    });
+
```
elimanzo, line 108: I think the instantaneous feedback for searching and displaying items is a bit jarring. My suggestion would be to add a delay / debouncing the search results as the user is typing. I have an example below, and this could apply to the other searches.

```Typescript

useEffect(() => {
      let delay = 500;
      if (searchQuery.length > 3) {
        delay = 300;
      }
      const timeOutId = setTimeout(() => {
        handleSearch();
      }, delay);

      return () => {
        clearTimeout(timeOutId);
      };
}, [data, handleSearch, searchQuery]);
```

`handleSearch` can be a void function that sets the data / searchResults, and will be similar code to what you provided.

elimanzo (changes requested):
Thanks for the PR! I have a quick suggestion below. 

I do believe that the searching / filtered feature will need to get refactored since it seems like we're making copies of all of our forms using `filter`. However, this is a full stack task for a different PR 😄

client/pages/Forms/Forms.tsx
abanuelo: Sure @elimanzo debouncing is an excellent add-on! Will work to get this in!

abanuelo:
> Thanks for the PR! I have a quick suggestion below.
> 
> I do believe that the searching / filtered feature will need to get refactored since it seems like we're making copies of all of our forms using `filter`. However, this is a full stack task for a different PR 😄

Added ticket to issues: #112

elimanzo (approved):
Looks good. TY

PR #116, opened 2024-08-19
Feat (form builder): Payment Component

elimanzo:
## Changes
1. Added Draggable Payment component
2. Added a util function to normalize payment amount (to adjust for Stripe limitations)
3. Added Translations

## Purpose

In preparation for the stripe component when we build the form based off of [figma](https://www.figma.com/design/0RPE1KFkTvkcdvWoYsrxhe/cascarita-logo-ideas?node-id=1515-1275&t=DCVRwbvWljCxKhow-0) designs

image

### Note
1. When the form is created, there won't be any component for this field (Will be done in another PR)
2. Builder Payment component still needs to be able to retrieve all accounts. (Currently blocked on this)
3. Will eventually need to create a drop down for the currency, but since we're sticking with USD for now, I jut left this as is

### DIFF
--- a/client/components/DragAndDropComponents/DNDCanvas/DNDCanvas.tsx
+++ b/client/components/DragAndDropComponents/DNDCanvas/DNDCanvas.tsx
@@ -16,6 +16,7 @@ import { v4 as uuidv4 } from "uuid";
 import DraggablePhoneNumber from "../DraggablePhoneNumber/DraggablePhoneNumber";
 import DraggableEmail from "../DraggableEmail/DraggableEmail";
 import { StrictModeDroppable } from "../../StrictModeDroppable/StrictModeDroppable";
+import DraggablePayment from "../DraggablePayment/DraggablePayment";
 
 const DNDCanvas = forwardRef(
   (
@@ -45,6 +46,7 @@ const DNDCanvas = forwardRef(
       long_text: DraggableLongText,
       email: DraggableEmail,
       phone_number: DraggablePhoneNumber,
+      payment: DraggablePayment,
     };
 
     const appendField = (item: DroppedItem) => {
@@ -91,6 +93,14 @@ const DNDCanvas = forwardRef(
           id: item.id,
           ref: item.id,
           properties: { default_country_code: "US" },
+          validations: { required: false },
+          type: item.type,
+        },
+        payment: {
+          title: "",
+          id: item.id,
+          ref: item.id,
+          properties: { currency: "USD" },
           validations: { required: false },
           type: item.type,
         },
--- a/client/components/DragAndDropComponents/DNDCanvas/DNDCanvas.tsx
+++ b/client/components/DragAndDropComponents/DNDCanvas/DNDCanvas.tsx
@@ -100,7 +100,10 @@ const DNDCanvas = forwardRef(
           title: "",
           id: item.id,
           ref: item.id,
-          properties: { currency: "USD" },
+          properties: {
+            price: { type: "fixed", value: "", currency: "USD" },
+            description: "",
+          },
           validations: { required: false },
           type: item.type,
         },
--- a/client/components/DragAndDropComponents/DraggablePayment/DraggablePayment.tsx
+++ b/client/components/DragAndDropComponents/DraggablePayment/DraggablePayment.tsx
@@ -72,7 +72,7 @@ const DraggablePayment: React.FC<DraggablePaymentProps> = ({
                 render={({ field }) => (
                   <input
                     {...field}
-                    placeholder={"Title"}
+                    placeholder={t("title")}
                     className={styles.question}
                   />
                 )}
@@ -85,7 +85,7 @@ const DraggablePayment: React.FC<DraggablePaymentProps> = ({
                   render={({ field }) => (
                     <input
                       {...field}
-                      placeholder={"Description"}
+                      placeholder={t("description")}
                       className={styles.question}
                     />
                   )}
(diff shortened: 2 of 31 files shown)


client/components/DragAndDropComponents/DNDCanvas/DNDCanvas.tsx
```diff
@@ -94,6 +96,17 @@ const DNDCanvas = forwardRef(
           validations: { required: false },
           type: item.type,
         },
+        payment: {
+          title: "",
+          id: item.id,
+          ref: item.id,
+          properties: {
+            price: { type: "fixed", value: "", currency: "USD" },
```
abanuelo, line 104: In the event that we support multiple currencies, can we have USD be an variable?

client/components/DragAndDropComponents/DraggablePayment/DraggablePayment.tsx
```diff
@@ -0,0 +1,174 @@
...
+    handleResize();
+    window.addEventListener("resize", handleResize);
+
+    return () => {
+      window.removeEventListener("resize", handleResize);
+    };
+  }, []);
+
+  const handleClick = () => {
+    setIsMenuOpen((prev) => !prev);
```
abanuelo, line 45: Out of curiosity why not just call `setIsMenuOpen(!prev)`?

client/components/DragAndDropComponents/DraggablePayment/DraggablePayment.tsx
```diff
@@ -0,0 +1,174 @@
...
+        <div
+          ref={provided.innerRef}
+          {...provided.draggableProps}
+          {...provided.dragHandleProps}
+          style={provided.draggableProps.style}
+          onClick={handleClick}>
+          <div style={{ position: "relative" }} ref={containerRef}>
+            <p className={styles.textElementTypeText}>{t("payment")}</p>
+            <div
+              style={{
```
abanuelo, line 60: Can we push all in-line styling to the relevant `*.module.css`? Better for readability and configurability later down the line!

client/components/DragAndDropComponents/DraggablePayment/DraggablePayment.tsx
```diff
@@ -0,0 +1,174 @@
...
+                    </div>
+                    <p className={styles.currency}>
+                      {properties.price?.currency}
+                    </p>
+                  </div>
+                </>
+              )}
+              {/* TODO: Add Stripe Accounts */}
+              <select className={styles.input}>
+                <option value="">Select an Account</option>
```
abanuelo, line 124: use a map here? I know this is going to change but just make it cleaner

client/pages/NewForm/NewForm.tsx
```diff
@@ -95,7 +96,7 @@ const NewForm = () => {
 
   // TODO: save by mongo form ID
   const onSave = async (data: Form) => {
-    if (formId == null || formId === undefined) {
+    if (formId == null) {
```
abanuelo, line 99: change to `!formId`

client/util/formatPayment.ts
```diff
@@ -0,0 +1,32 @@
+// Method to format payment based off of Stripe limitations
+
+export const formatPayment = (value: string) => {
```
abanuelo, line 3: I am generally against having oneoff util files unless absolutely necessary. I would say for this case, let's ahead and couple this to where it is being applied within the `DraggablePayment` component?

abanuelo (changes requested):
Overall very good work and great scaffolding so that we can plug and play new features as they get released into the code base. A couple of nits that I would advise you to take care of so that we can get the ball moving!

client/components/DragAndDropComponents/DNDCanvas/DNDCanvas.tsx
elimanzo: Hey @abanuelo can you provide a bit of clarity here. I used `currency` as the variable that could be changed later on.

client/components/DragAndDropComponents/DraggablePayment/DraggablePayment.tsx
elimanzo: ack!

client/components/DragAndDropComponents/DraggablePayment/DraggablePayment.tsx
elimanzo: ack!

client/util/formatPayment.ts
elimanzo: > I am generally against having oneoff util files unless absolutely necessary. I would say for this case, let's ahead and couple this to where it is being applied within the `DraggablePayment` component?

So, I mentioned this in the a standup, but there's a possibility to allow the user to input their own amount if the form had the stripe as some optional donation, so this function would be used for that case, and we can avoid coupling. 

As for having one off util, I prefer breaking up logic like this into separate files / function, it makes components a bit more easier to read.

If you don't see this being used in the future I could add to `DraggablePayment` component

client/components/DragAndDropComponents/DNDCanvas/DNDCanvas.tsx
abanuelo: My rationale here @elimanzo is to eventually have currency tied to some of our stripe tables so we can keep track of internal auditing. This being hardcoded for now is totally fine but just as we consider scaling it might be a nice to have. I will accept as is for now!

client/pages/NewForm/NewForm.tsx
abanuelo: Ah gotcha! It's a catch all! Thanks for the thorough explanation!

client/util/formatPayment.ts
abanuelo: Let's keep for now but consider dragging into the component later on!

abanuelo (approved):
LGTM! Thanks for being thorough with the changes and responding to feedback as needed!

client/components/DragAndDropComponents/DNDCanvas/DNDCanvas.tsx
elimanzo: I'm still a little confused haha, but maybe in a meeting we can discuss it a bit more 😃 .

PR #117, opened 2024-08-29
Feature/bread crumbs final

jipatino98:
### Description
This PR introduces a dynamic breadcrumb component that navigates users to previous pages as they navigate the dashboard.

**Note**:  Migrating to the newest router of the React Router library, had us made changes to the `AuthContext.tsx` file. More specifically, we had to remove the `useNavigate` hook because it was being used outside the scope of the Router context so it caused issues. One solution I saw on stack overflow was to save the token into `localStorage`. However, I am not sure of the security risk this may pose. 

Special Thanks to @abanuelo for helping me overcome a pesky bug I was facing with React-Router's `<Outlet />` component. 

https://github.com/user-attachments/assets/e92b5454-3f3e-4dbf-a4dd-181d22f75f28

### Checklist


abanuelo (approved):
LGTM but can we make a ticket so that we can update the Auth Context? Good work!

PR #118, opened 2024-09-12
feat: Updated UI for Register

jipatino98:
### Description

Made changes to the Register UI

### Issue Link

N/A

### Checklist

### DIFF
--- a/client/api/users/service.ts
+++ b/client/api/users/service.ts
@@ -35,7 +35,7 @@ const registerUser = async (
   city: string | null,
   state: string | null,
   zipCode: string | null,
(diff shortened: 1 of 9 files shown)


client/api/users/service.ts
```diff
@@ -63,6 +63,7 @@ const registerUser = async (
     });
 
     const result = await response.json();
+    console.log(result);
```
abanuelo, line 66: Nit: clean up console logs as needed

abanuelo (approved):
Looks great other than some minor clean up work! @jipatino98 @aplprz19 correct those changes first, but will approve on my end!

PR #119, opened 2024-09-21
feat: integrate Auth0 service

abanuelo:
### Description

Working to integrate Auth0 service directly into cascarita to handle (a) session management, (b) authn and (c) authz, although this is a longer requested change we will implement in future versions of our app! Goal is to have all authentication related needs managed by Auth0 so we can focus on fast development and iteration! 

Below are videos of Auth0 creating a persistent session for the user along with opening up the registration modal if not currently registered with Cascarita!

### Existing group registration flow

Recording 2024-10-19 at 14 47 57

### New Group registration flow
Recording 2024-10-19 at 14 50 07

### Checklist


client/api/users/service.ts
```diff
@@ -22,44 +22,25 @@ const updateUsersLanguages = async (
...
-      zip_code: zipCode,
-      // logo_url: logoUrl,
+    const formData = {
+      group_id: data.group_id,
+      name: data.name,
+      streetAddress: data.streetAddress,
+      city: data.city,
+      state: data.state,
+      zipCode: data.zipCode,
+      logoUrl: data.logoUrl,
```
elimanzo, line 34: I believe we can [spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) here fort his object

```suggestion
      ...data
```

client/auth_config.json
```diff
@@ -0,0 +1,5 @@
+{
+    "domain": "dev-2vszya8j41e1n3fe.us.auth0.com",
+    "clientId": "Bfvu7AdllYYt0ONhhRILk9xaUr4qeDVu",
+    "audience": "https://dev-2vszya8j41e1n3fe.us.auth0.com/api/v2/"
```
elimanzo, line 4: Just a quick question. These aren't secrets we need to hide right?

client/components/Forms/DivisionForm/DivisionForm.tsx
```diff
@@ -23,7 +23,8 @@ const DivisionForm: React.FC<DivisionFormProps> = ({
 }) => {
   const [divisionName, setDivisionName] = React.useState("");
   const [isLoading, setIsLoading] = React.useState(false);
-  const { currentUser } = useAuth();
+  const { user } = useAuth0();
+  const currentUser = user;
```
elimanzo, line 27: I don't think  we need another intermediate variable for `user` Maybe just change everything that depends on `currentUser` to just `user`

client/components/Forms/LeagueForm/LeagueForm.tsx
```diff
@@ -23,7 +23,8 @@ const LeagueForm: React.FC<LeagueFormProps> = ({
   const [leagueName, setLeagueName] = React.useState("");
   const [leagueDesc, setLeagueDesc] = React.useState("");
 
-  const { currentUser } = useAuth();
+  const { user } = useAuth0();
+  const currentUser = user;
```
elimanzo, line 27: same comment as above :)

client/components/RegistrationModal/RegistrationModal.tsx
```diff
@@ -0,0 +1,232 @@
...
+  onRegistrationComplete,
+}) => {
+  const [page, setPage] = useState<number>(1);
+  const [isExistingOrg, setisExistingOrg] = useState<string>("No");
+  const [org, setOrg] = useState<string>("");
+  const [address, setAddress] = useState<string>("");
+  const [city, setCity] = useState<string>("");
+  const [state, setState] = useState<string>("");
+  const [zipCode, setZipCode] = useState<string>("");
+  const [selectedOrg, setSelectedOrg] = useState<string>("");
```
elimanzo, line 34: TS can already infer its type so it won't be needed. Also since  this logic is dependent on submitting a form consider using [useReducer](https://react.dev/reference/react/useReducer)


```suggestion
  const [page, setPage] = useState(1);
  const [isExistingOrg, setisExistingOrg] = useState("No");
  const [org, setOrg] = useState("");
  const [address, setAddress] = useState("");
  const [city, setCity] = useState("");
  const [state, setState] = useState("");
  const [zipCode, setZipCode] = useState("");
  const [selectedOrg, setSelectedOrg] = useState("");
```

client/components/SelectMenu/SelectMenu.tsx
```diff
@@ -61,7 +61,7 @@ const SelectMenu: React.FC<SelectMenuProps> & {
       value={value}
       name={name}
       onValueChange={onValueChange}
-      required={required}>
+      required={required ? true : false}>
```
elimanzo, line 64: I would recommend to change this variable to `isRequired,` and using [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)
```suggestion
      required={required ?? false}>
```

client/index.tsx
```diff
@@ -1,17 +1,67 @@
...
+  history.push(
+    appState && appState.returnTo
+      ? appState.returnTo
+      : window.location.pathname,
+  );
+};
+
+// Please see https://auth0.github.io/auth0-react/interfaces/Auth0ProviderOptions.html
+// for a full list of the available properties on the provider
+const config = getConfig();
```
elimanzo, line 47: is the `getConfig()` being used anywhere else in the code base? if not we can remove the export, and a super pro move is to make it to an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE). if we are then we can leave it as is :)

client/pages/Home/Home.tsx
```diff
@@ -1,27 +1,68 @@
...
+        try {
+          Cookies.set("email", user.email || "");
+          const token = await getAccessTokenSilently();
+          const response = await fetchUser(user.email || "", token);
+
+          if (response.isSigningUp) {
+            setRegistered(false);
+          } else {
+            setRegistered(true);
+          }
```
elimanzo, line 26:
```suggestion
          setRegistered(response?.isSigningUp ?? false);

```

client/pages/Leagues/Leagues.tsx
```diff
@@ -31,9 +30,9 @@ const Leagues = () => {
   const [isEditOpen, setIsEditOpen] = useState(false);
   const [isDeleteOpen, setIsDeleteOpen] = useState(false);
 
-  const { currentUser } = useAuth();
+  // const currentUser = {};
```
elimanzo, line 33: we can  remove this if we don't need it :)

client/pages/NewForm/NewForm.tsx
```diff
@@ -37,7 +40,18 @@ const NewForm = () => {
...
+
+  const emptyUser = {
+    id: 1,
+    email: "t@abc.com",
+    first_name: "string",
+    last_name: "string",
+    group_id: 1,
+    role_id: 1,
+    language_id: 1,
+  } as User;
```
elimanzo, line 54: little confused on why we need an `emptyUser` here and reassigning `currentUser`?

server/controllers/group.controller.js
```diff
@@ -23,6 +23,15 @@ const GroupController = function () {
     }
   };
 
+  var getAllGroups = async function (req, res, next) {
```
elimanzo, line 26: I never really understood why our server uses `var`. We should start using `let` / `const`. But this can be done in a different PR to refactor this, since it's all over the servers codebase.

server/utilityFunctions/auth0.js
```diff
@@ -0,0 +1,21 @@
+async function getUserInfoFromAuth0(token) {
+  try {
+    const response = await fetch(
+      "https://dev-2vszya8j41e1n3fe.us.auth0.com/userinfo",
```
elimanzo, line 4: is this a constant that we're always going to be using? Maybe make it to an env variable?

elimanzo (changes requested):
Thanks for the PR! I have a few comments about some of these changes, and I think it's good to go.

client/auth_config.json
abanuelo: Recommended action to make these environment variables but no they are not secrets. See thread here for details: https://community.auth0.com/t/safe-management-of-auth0-information-in-source-control/19051/3

client/api/users/service.ts
abanuelo: Perfect use of spread actually! Thank you!

client/components/Forms/DivisionForm/DivisionForm.tsx
abanuelo: Done! On top of that we don't want to use the `user` from the `useAuth0` since these user types are Auth0's user type. We have our own concept of User where `group_id` is in our schema (as you are aware!). I will document this change but this PR assumes Auth0 will handle tracking user's email, picture, roles, etc but the internal business logic (ie group_id user language preference), is all handled by us.

client/components/Forms/LeagueForm/LeagueForm.tsx
abanuelo: Applied changes!

client/components/SelectMenu/SelectMenu.tsx
abanuelo: Done! Thanks for teaching me something new!

client/components/RegistrationModal/RegistrationModal.tsx
abanuelo: Done! Thank you for the knowledge! Also implemented `useReducer` alternative to form ingestion!

client/pages/Home/Home.tsx
abanuelo: Did not use nullish coalescing here since there is some backend logic we need to implement. Reason being: If the `isSigningUp` key is in the response body, it will assume its value as `true` which in case sets the `setRegistered` state to true which hides the modal. There are only two values for this state at the moment (true or undefined). I think for now we need to consider how we can reimplement this but for the time being will optimize it with ternary operator!

client/pages/Leagues/Leagues.tsx
abanuelo: Done! But I am glad I commented this out and not removed because we actually want to be able to reference the current logged in user and fetch their group_id so that when Leagues are created with the LeagueModal, they can be properly seen and displayed them on the table!

client/pages/NewForm/NewForm.tsx
abanuelo: Temporary changes needed to resolve some errors when debugging auth0! Thanks for catching, I have removed this `emptyUser`.

server/utilityFunctions/auth0.js
abanuelo: On it!

server/controllers/group.controller.js
abanuelo: Agreed! Issued created here: https://github.com/cascaritaco/cascarita/issues/132

client/index.tsx
abanuelo: Done! Made this into an IIFE and removed the export!

client/components/RegistrationModal/RegistrationModal.tsx
```diff
@@ -0,0 +1,258 @@
...
+                    <SelectMenu.Item key={group.id} value={group.id.toString()}>
+                      {group.name}
+                    </SelectMenu.Item>
+                  ))}
+                </SelectMenu>
+              )}
+            </div>
+
+            {state.isExistingOrg === "Yes" ? (
+              <button style={{ marginTop: "26px" }}>Finish</button>
```
elimanzo, line 157: Not sure if you had a preference to use `className` vs inline `style` for this button margin

client/pages/Leagues/Leagues.tsx
```diff
@@ -25,15 +27,24 @@ const Leagues = () => {
   const [searchQuery, setSearchQuery] = useState("");
   const [debouncedQuery, setDebouncedQuery] = useState("");
 
+  const { getAccessTokenSilently } = useAuth0();
+  const [groupId, setGroupId] = useState(null);
```
elimanzo, line 31: I  believe here we do need to set a type because in line 44, it will error if it sets anything other than null.

```suggestion
  const [groupId, setGroupId] = useState<number | null>(null);
```

client/components/Forms/LeagueForm/LeagueForm.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
```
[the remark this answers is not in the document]
jipatino98: Yeah, I can fix that! I'll get that addressed asap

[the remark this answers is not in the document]
abanuelo: @jipatino98 I will leave this to you!

client/components/RegistrationModal/types.ts
```diff
@@ -0,0 +1,16 @@
+export interface State {
+  page: number;
+  isExistingOrg: string;
```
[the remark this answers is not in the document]
abanuelo: Done!

client/components/RegistrationModal/types.ts
```diff
@@ -0,0 +1,16 @@
...
+  org: string;
+  address: string;
+  city: string;
+  state: string;
+  zipCode: string;
+  selectedOrg: string;
+}
+
+export type Action =
+  | { type: "SET_FIELD"; field: keyof State; value: string | number }
```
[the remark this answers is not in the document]
abanuelo: Done!

client/pages/Leagues/Leagues.tsx
abanuelo: Done!

client/components/RegistrationModal/RegistrationModal.tsx
abanuelo: Done!

PR #121, opened 2024-09-26
Feature/settings page

jipatino98:
#### Main Task
- Created the UI for the Plan and Payment tabs in the Settings page
- Created a reusable component for the Navbar, that will allow us to place navigational tabs on our dashboard. In order for this to work properly, make sure that the each tab corresponds a route that is nested within a common parent route. Will add to the Component Docs
- Created a reusable component for a ProgressBar,  that will allow us to measure the current value compared to a max value for things such as storage, players, etc. Will add to the Component Docs

#### Minor Tweaks
- Fixed up some css properties for existing Components
- Created a directory for custom React hooks. I have created a hook, `useFormatDate(date: number) : string`  for Date conversion into a human readable string. 

#### Note
- We still need to connect the Stripe Account Form to endpoints from the server. This will be another PR. 

### Issue Link

N/A

### Checklist


client/pages/Routes.tsx
```diff
@@ -104,7 +106,12 @@ export const useRouter = () =>
         />
         <Route path="forms" element={<Forms />} />
         <Route path="forms/check" element={<NewForm />} />
-        <Route path="settings" element={<Settings />} />
+        <Route path="settings" element={<Settings />}>
+          <Route index element={<Plan />} />
+          <Route path="payment" element={<Payment />} />
+          {/* <Route path="billing" element={<Billing />} /> */}
```
abanuelo, line 112: nit: remove

server/controllers/account.controller.js
```diff
@@ -112,6 +112,54 @@ const AccountController = function () {
...
+      let account;
+      let accountId;
+
+      const allExistingStripeAccounts = await UserStripeAccounts.findAll({
+        where: {
+          group_id: user.group_id,
+        },
+      });
+
+      // if (allExistingStripeAccounts) {
```
abanuelo, line 127: Add comment as to why this commented out!

abanuelo (changes requested):
Some minor changes but everything else should be good!

server/controllers/account.controller.js
jipatino98: Okay just added the comment, the reason it was commented out rather then removed is because we are still figuring what logic we need to return all existing accounts and have them link to specific stripe accounts.

abanuelo (approved):
LGTM!

PR #122, opened 2024-09-26
Add workflow to push images to ECR registry

garcialuis:
### Description
This PR includes two changes:
- Adds caching for npm node modules to main ci, this will speed up workflow run time when there are cache hits.
- Adds workflow to build and push images to ECR registry when there are updates made to the main branch

### Additional Notes
Corresponding secrets have been added to the repo and private ECR repositories have been created for the images.


abanuelo (approved):
LGTM! Thanks for adding this workflow file

abanuelo:
One other thing, can we change the trigger for this workflow build? Instead of all merges into `main`, can we create a new branch called deploy and manually set up a process where we merge main into deploy and then that will trigger image build and push to ECR?

garcialuis:
> One other thing, can we change the trigger for this workflow build? Instead of all merges into `main`, can we create a new branch called deploy and manually set up a process where we merge main into deploy and then that will trigger image build and push to ECR?

That sounds reasonable. A downside I see from that is that we will continue to have a manual process to build new images. By building a new image every time code is pushed to main we can get in the habit of making sure that every change is production ready. This will also help us continue to move towards a continuous deployment strategy. 

What we should do to accomplish what you're suggesting is having a `staging` branch. Our staging branch can build our images and deploy to a staging environment where we can run additional tests before merging to main.
Once those additional tests are completed we can merge into main, which can also deploy the changes to production.

PR #123, opened 2024-09-30
Feat/add stripe status

saulRM10:
### Checklist

### DIFF
--- a/server/controllers/account.controller.js
+++ b/server/controllers/account.controller.js
@@ -2,7 +2,12 @@
 
 require("dotenv").config();
 const Stripe = require("stripe")(process.env.STRIPE_TEST_API_KEY);
-const { UserStripeAccounts, FormPaymentIntents, User } = require("../models");
+const {
+  UserStripeAccounts,
+  FormPaymentIntents,
+  User,
+  StripeStatus,
+} = require("../models");
 const { where } = require("../mongoModels/response");
 
 const AccountController = function () {
@@ -38,6 +43,11 @@ const AccountController = function () {
       });
 
       if (!existingStripeAccount) {
+        const defaultStatus = await StripeStatus.findOne({
+          where: {
+            status: "Restricted",
+          },
+        });
         await UserStripeAccounts.create({
           user_id: user.id,
           stripe_account_id: accountId,
@@ -45,6 +55,10 @@ const AccountController = function () {
           requires_verification: true,
           charges_enabled: false,
           payouts_enabled: false,
+          platform_account_name: user.platform_account_name,
+          platform_account_description: user.platform_account_description,
+          account_email: user.account_email,
+          stripe_status_id: defaultStatus.id,
         });
       }
 
--- a/server/controllers/account.controller.js
+++ b/server/controllers/account.controller.js
@@ -2,7 +2,8 @@
 
 require("dotenv").config();
 const Stripe = require("stripe")(process.env.STRIPE_TEST_API_KEY);
-const { UserStripeAccounts, FormPaymentIntents } = require("../models");
+const { UserStripeAccounts, FormPaymentIntents, User } = require("../models");
+const { where } = require("../mongoModels/response");
 
 const AccountController = function () {
   var createAccountConnection = async function (req, res, next) {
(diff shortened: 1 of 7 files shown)


server/controllers/account.controller.js
```diff
@@ -112,11 +126,85 @@ const AccountController = function () {
     }
   };
 
+  var getAllAccountsByGroupId = async function (req, res, next) {
+    try {
+      const groupId = req.params.id;
```
garcialuis, line 131: We should add handling for groupId being undefined

server/controllers/account.controller.js
```diff
@@ -112,11 +126,85 @@ const AccountController = function () {
...
+      let data = accounts.length != 0 ? accounts : [];
+
+      res.status(200).json(data);
+    } catch (error) {
+      next(error);
+    }
+  };
+
+  var calculateStripeStatus = async function (account) {
+    let rejectedReasons = [
```
garcialuis, line 164: These are not intended to change, so lets declare it as a constant. Constant arrays are still mutable but at least it will be clear to other devs

garcialuis (changes requested):
Overall changes look good, I left two comments to address.

server/controllers/account.controller.js
saulRM10: Good catch, @villaleo  made this cool utility function to validate the id of a given model.

server/controllers/account.controller.js
saulRM10: fixed as well, thanks

server/controllers/form.controller.js
```diff
@@ -92,9 +92,17 @@ const FormController = {
 
   async getFormByDocumentId(req, res, next) {
```
saulRM10, line 93: @elimanzo are you opposed to this ? I am adding another query to this route to then append it with the `sql_form_id` ? Do you want a separate route ?

elimanzo: this should be good

server/controllers/form.controller.js
```diff
@@ -92,9 +92,17 @@ const FormController = {
...
+      let resultsObject = results.toObject();
+      const sqlFormData = await Form.findOne({
+        where: {
+          document_id: form_document_id,
+        },
+      });
 
-      return res.status(201).json(results);
+      resultsObject.sql_form_id = sqlFormData.id;
+      return res.status(201).json(resultsObject);
```
garcialuis, line 105: A return status code of 200 (OK) would be more suitable for this handler.

server/controllers/form.controller.js
```diff
@@ -92,9 +92,17 @@ const FormController = {
 
   async getFormByDocumentId(req, res, next) {
     try {
-      const results = await FormMongo.findById(req.params.document_id);
+      const form_document_id = req.params.document_id;
+      let results = await FormMongo.findById(form_document_id);
+      let resultsObject = results.toObject();
```
garcialuis, line 97: Here for `results` as well as `sqlFormData` in line 98 we are assuming that a record will be retrieved from the db. We should check if they're null before trying to convert to an obj or assigning new values to it.

server/controllers/form.controller.js
```diff
@@ -92,9 +92,17 @@ const FormController = {
...
+        where: {
+          document_id: form_document_id,
+        },
+      });
 
-      return res.status(201).json(results);
+      resultsObject.sql_form_id = sqlFormData.id;
+      return res.status(201).json(resultsObject);
     } catch (error) {
       next(error);
```
garcialuis, line 107: We should log the error before passing it over to the next middleware.

server/controllers/account.controller.js
```diff
@@ -112,11 +129,113 @@ const AccountController = function () {
...
+    const stripeStatusId = await StripeStatus.findOne({
+      where: {
+        status: status,
+      },
+    });
+
+    return stripeStatusId.id;
+  };
+
+  const getPulishableKey = function (req, res, next) {
```
garcialuis, line 227: What's the purpose for this route? And where is this key set? I don't see it in the .example.env file
There's a typo in the env var name: STRIPE is misspelled

garcialuis (changes requested):

garcialuis (approved):
LGTM

PR #131, opened 2024-10-20
doc: swagger api doc integration

villaleo:
# Add Swagger API Documentation

## Overview
This PR introduces Swagger API documentation to our project, enhancing our development workflow and API testing capabilities.

## Key Benefits
- **Interactive API Testing**: Developers can now test endpoints directly within the project interface.
- **Improved Developer Experience**: Clear, accessible documentation for all API endpoints.
- **Faster Onboarding**: New team members can quickly understand and interact with our API structure.
- **Client Integration Support**: Easier for frontend developers and external partners to integrate with our API.

## Implementation Details
- Added Swagger JSDoc comments to model files in `server/models/`
- Created Swagger route documentation for `account.routes.js`
- Integrated Swagger UI for visual representation of our API

## Important Consideration
To maintain the value of this documentation:
- **Keep Swagger Docs Updated**: When business decisions lead to API changes, it's crucial to update the corresponding Swagger documentation.
- **Review Process**: PR reviews should include checking for updates to Swagger docs when API changes are made.

## Next Steps
- Add Swagger documentation for remaining route files
- Consider automating parts of the documentation process

## Testing
- Access Swagger UI at `/api-docs` to view and test the documented endpoints

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -68,6 +68,8 @@
         "stripe": "^16.0.0",
         "style-loader": "^3.3.3",
         "supertest": "^6.3.3",
+        "swagger-jsdoc": "^6.2.8",
+        "swagger-ui-express": "^5.0.1",
         "ts-loader": "^9.5.1",
         "typescript": "^5.3.3",
         "url-loader": "^4.1.1",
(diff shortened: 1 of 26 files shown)


abanuelo (changes requested):
Super clean! Thanks for getting this in @villaleo one minor suggestion is can we make this endpoint `/api/docs` instead of `/api-docs`? Also is there a way without directly adding comments above each command to be able to specify an endpoints behavior. Like for instance, can we do it directly from the code logic itself?

villaleo:
> Super clean! Thanks for getting this in @villaleo one minor suggestion is can we make this endpoint `/api/docs` instead of `/api-docs`? Also is there a way without directly adding comments above each command to be able to specify an endpoints behavior. Like for instance, can we do it directly from the code logic itself?

I made the changes to route from `/api-docs` to `/api/docs`. I like your suggestion of having the docs in a different location! I did some research but it looks like this is a limitation of Swagger. I will do more research and see if we can build the docs from the code logic. I found someone created a python script to do this, but it seems sus since it has zero stars and was created by some random person.

PR #133, opened 2024-10-22
feat: add AWS SDK to create .env file

abanuelo:
### Description

As our `.env` grows, it will be best practice to keep them somewhere secure and pull them down for development. I have created a `create-env.js` file that fetches from AWS Secrets Manager to create the `.env` file. You will need to setup AWS account on your local machine with profile name `cascarita` to be able to fetch these credentials. 

### Checklist


create-env.js
```diff
@@ -0,0 +1,52 @@
...
+        .join("\n");
+
+      fs.writeFileSync(envFilePath, envFileContent);
+
+      console.log(
+        `.env file created at ${envFilePath} with secrets from AWS Secrets Manager.`,
+      );
+    } else {
+      throw new Error("SecretString is empty");
+    }
```
elimanzo, line 43: I think removing the else is a bit more concise here.

```suggestion
    if (!secretString) {
      throw new Error("SecretString is Empty");
    }
    
    const secretObject = JSON.parse(secretString);

    const envFileContent = Object.entries(secretObject)
      .map(([key, value]) => `${key}=${value}`)
      .join("\n");

    fs.writeFileSync(envFilePath, envFileContent);

    console.log(
      `.env file created at ${envFilePath} with secrets from AWS Secrets Manager.`,
    );
```

elimanzo (approved):
Looks really good, just 1 small nit 😄

abanuelo:
Note going to hold off on merging this in as it requires some documentation on how to configure the AWS credentials locally so that you can leverage the SDKs in the script! Will work with @villaleo to get some of this documentation in!

PR #142, opened 2024-10-27
fix: docker image for terraform deployments

abanuelo:
### Description

As @garcialuis worked on integrate Terraform, there were some problems with the deployment of the container to ECS. This PR will only allow us to run the deployment of the `cascarita-server` image to our private registry. This image will have server backend code with the frontend exposed via the `bundle.js`. This is a temporary deployment fix until we develop a reverse proxy solution!

### Checklist


abanuelo:
Note there is still work to add certificate manager to elastic load balancer to be able to redirect https traffic to get Auth0 working inside terraform will create an issue for this

terraform/ec2.tf
```diff
@@ -36,11 +36,40 @@ resource "aws_iam_role" "ecs_task_execution_role" {
...
+#         Resource = "*"
+#       }
+#     ]
+#   })
+# }
+
 resource "aws_iam_role_policy_attachment" "ecs_task_execution_policy" {
   role       = aws_iam_role.ecs_task_execution_role.name
-  policy_arn = var.task_policy_arn
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"
```
garcialuis, line 65: We should either use the task_policy_arn variable here or get rid of the variable as it will no longer be used.

abanuelo: Ack! Implemented!

abanuelo:
Deployment is fully functional: https://cascarita.io/

Issue #141, opened 2024-10-24
[Enhancement] Build out Users Page


abanuelo:
Goal is to build out the remaining parts of the users page
<img width="1858" alt="Screenshot 2024-10-24 at 12 17 38 AM" src="https://github.com/user-attachments/assets/551b2b0e-a1fd-4b06-ad83-d020d3dc06fa">

abanuelo:
Feel free to tag your PR to this issue @alonso-ov !

alonso-ov:
[Pull Request #146](https://github.com/cascaritaco/cascarita/pull/146)

@abanuelo 👍👍👍!

PR #146, opened 2024-11-04
Feat/add users page

alonso-ov:
### Description

This pull request adds functionality to the user management tab. Clients can add, edit, filter, and remove users.

### Issue Link

[Issue #141](https://github.com/cascaritaco/cascarita/issues/141)

### Checklist

### Additional Notes

Some things to note in this pull requests is the following:

- Before creating a user, the program checks that there are no duplicate emails within the group using the `group_id` attribute. This means that an email can only appear once within a group but can appear multiple times in the database by appearing in multiple groups. If the email already exists in the group, the user will not be created, and the client will be informed that the email already exists within the group.

- When a new user is created, a random password is generated. This random password still needs to be sent out to the provided email.

- When deleting a user, the deletion happens in the database, but the page needs to be refreshed to see the user's card disappear. A forced refresh is needed.


abanuelo (approved):
Looks great! Added some minor changes on my end! I think the addUser functionality might be broken but we can address this in separate PR will keep you posted @alonso-ov

Issue #147, opened 2024-11-05
[Bug] CRUD calls for User page on app not fully functional


abanuelo:
As part of #141, we decided to merge in the existing code and just create a couple of PRs relevant to update the logic for the CRUD calls in the `/users` route. Assigning Alonso to take a look.

PR #148, opened 2024-11-05
fix: add user by removing password as handled by Auth0

abanuelo:
### Description

Noticed that `bycrypt` was being used for password generation. All passwords and role handling should go through Auth0 so we should be okay to remove that dependency as needed.

### Issue Link

#147 

### Checklist


elimanzo (approved):
LGTM! I'm assuming `bcrypt` was deleted in another commit?

Issue #135, opened 2024-10-23
[Bug] Drill down into Division breaks app


abanuelo:
As QAing our current application I noticed that when we drill into a division we hit an error. Assigning this ticket to @jipatino98 as he is the owner of this breadcrumb flow:
![Recording 2024-10-22 at 21 43 14](https://github.com/user-attachments/assets/8b41cb4f-ba70-46a7-a9db-e952398fd5e8)

jipatino98:
addressed issue! Check [Bug/seasons breadcrumb fix #152](https://github.com/cascaritaco/cascarita/pull/152)

abanuelo:
Closing as no longer applicable from changes in #183 to linearize home experience

PR #152, opened 2024-11-10
Bug/seasons breadcrumb fix

jipatino98:
### Description

Addressed the issue that was brought up by @abanuelo  [[Bug] Drill down into Division breaks app #135].(https://github.com/cascaritaco/cascarita/issues/135)

The issue arose because we were using hooks (ie `useTranslation()`) inside the crumb property inside the routes in `Routes.tsx` that get called during the render of the Breadcrumbs component. This lead to inconsistent hook ordering when the user navigated to another page.

The solution was to modify `BreadCrumb.tsx` so that the crumb was defined as a React component rather than a function that returns a React Component. Moreover, I created a new file called `BreadCrumbComponents.tsx` that will house our individual BreadCrumb component that we can call instead in our Route definitions. This not only fixed the issue but also condensed our Route file.

I also addressed the issue where if a user creates a new League, Season, etc. and they add a `/` character to the name input field, it would redirect the user to a nonexistent page. For the mean time, the user is unable to add that character to prevent the error. In the future, we may need to change how we pass the name to URL to allow users to add the `/` to their names.

### Issue Link

N/A

### Checklist


client/components/BreadCrumb/BreadCrumb.tsx
```diff
@@ -11,22 +11,19 @@ interface IMatches {
...
-      const crumb = (match.handle as HandleType).crumb(
-        match.data as string | undefined,
-      );
-      return crumb as React.ReactNode;
-    });
+  const crumbs = useMemo(() => {
+    return matches
+      .filter((match) =>
+        Boolean(match.handle && (match.handle as HandleType).crumb),
+      )
```
elimanzo, line 24: I don't think we need the `Boolean` function if the expression can return truthy

```suggestion
      .filter(
        (match) => match.handle != null && (match.handle as HandleType).crumb,
      )
```
We may be able to remove our `IMatches` interface altogether though if we assert UIMatch annotations 

```TypeScript
  const matches = useMatches() as UIMatch<unknown, HandleType>[];

  const crumbs = useMemo(() => {
    return matches
      .filter((match) => match.handle != null && match.handle.crumb)
      .map((match) => match.handle.crumb);
  }, [matches]);
  ```

elimanzo (approved):
Looks good! I did have 1 nit, but not sure the right direction to take with the type assertions for `IMatches` to be more type safe

client/components/BreadCrumb/BreadCrumb.tsx
jipatino98: Nice catch, I removed those the `Boolean` function and used the `UIMatch` type instead.

elimanzo (approved):
LGTM!

PR #158, opened 2024-11-21
Feat/Create Junction Table UserRoles

JesusGo115:
### Description

This pull request will add a new junction table, UserRoles, that will allow users to have multiple roles using the user ID

### Checklist


abanuelo (approved):
🚀 Thanks for making these changes!

server/migrations/20241120060820-update-user-table-remove-role-id-column.js
```diff
@@ -0,0 +1,39 @@
...
+      throw error;
+    }
+  },
+
+  async down(queryInterface, Sequelize) {
+    const transaction = await queryInterface.sequelize.transaction();
+    try {
+      await queryInterface.addColumn("Users", "role_id", {
+        type: Sequelize.INTEGER,
+        allowNull: false,
```
saulRM10, line 23: This might break, it add back the `role_id` column but `allowNull` is still false. So youd' need to add a default `role_id` so it doesn't break and shoot the sequalize validation error. Id recommend just leaving it as `allowNull: true`

saulRM10 (changes requested):
Found a potential issue on the `down()` of one of your migrations. Quick fix tho. Besides that, great job chuy !

server/migrations/20241120060820-update-user-table-remove-role-id-column.js
JesusGo115: Implemented the change. Thanks for spotting that!

saulRM10 (commented):
Thanks for the fix, we have a user (Maria, from salinas soccer femenil) I think in the migration we need to assign any user in our db to `admin` so you will need to add a part in your migration that adds entries to the `UserRoles` table. What do you think @garcialuis @abanuelo

JesusGo115:
We could create a migration that sets up a UserRole row as "admin" for every existing user in the User table

saulRM10:
> We could create a migration that sets up a UserRole row as "admin" for every existing user in the User table

Can you add a migration for this on this pr ?

JesusGo115:
No

JesusGo115:
Jk

Yes, I added it. Please take a look!

saulRM10 (approved):
Looks good Chuy, thanks for adding the migrations that updates the users

PR #161, opened 2024-12-11
fix: adding copyright for software

abanuelo:
### Description

Adding copyright descriptions to README for non-commercial use and only allowed to use the software for personal or academic uses! Also any reference to our software needs to attribute cascarita.io

### Checklist


garcialuis (approved):
Year was updated, LGTM

PR #166, opened 2025-02-03
Fix: Use fixed pnpm version

garcialuis:
### Description

Corepack is used in the client and ran in one of the image build steps. This PR ensures that corepack does not use it's default pnpm version but instead use a fixed version. 

This update sets the pnpm version to `9.14.2`

### Additional Notes
This fixes image builds for the client container.


abanuelo (approved):
https://pnpm.io/docker says `pnpm` does not support local system volume mounts will change in the future

PR #167, opened 2025-02-03
Migrations for Roles and Permissions tables

garcialuis:
### Description

This PR is for [CV 35](https://cascarita.atlassian.net/browse/CV-35?atlOrigin=eyJpIjoiZTUwYzNkYzdlZDhiNDRjMjllNmM4ODM2YjJlYzkyMWQiLCJwIjoiaiJ9). It includes:
- migrations to create the Permissions and RolesPermissions tables.
- migrations to update Roles and UserRoles tables
- removes code that was no longer used for Roles.

### Issue Link

[CV 35](https://cascarita.atlassian.net/browse/CV-35?atlOrigin=eyJpIjoiZTUwYzNkYzdlZDhiNDRjMjllNmM4ODM2YjJlYzkyMWQiLCJwIjoiaiJ9)

### Additional Notes
This PR sets up our database tables for Role Based Access Control. 
Future steps towards that are:
- migration to seed known Permissions and Roles
- logic implementation to enforce authorization


abanuelo (commented):
Will defer final approval changes to @saulRM10 but [we created this ticket Jira ticket](https://cascarita.atlassian.net/jira/software/c/projects/CV/boards/4/backlog?selectedIssue=CV-57)

which will make some changes to the `UserRole` table potentially to be able to map the `UserStripeAccount` table back to a `User`.

server/models/role.js
```diff
@@ -8,26 +8,43 @@ module.exports = (sequelize, DataTypes) => {
      * The `models/index` file will call this method automatically.
      */
     static associate(models) {
-      Role.hasMany(models.UserRoles, {
-        foreignKey: "role_id",
-        sourceKey: "id",
-      });
+      // define association here
```
saulRM10, line 11: This makes sense not to have, bc we can query anything we need through RolePermission model.

saulRM10 (changes requested):
Just left one comment that needs addressing on the Permissions model. We should add an explicit association on the Permission model, so we can query both ways.

server/models/permissions.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+const { Model } = require('sequelize');
+
+module.exports = (sequelize, DataTypes) => {
+  class Permission extends Model {
+      static associate(models) {
+        //associations
```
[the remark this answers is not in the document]
garcialuis: @saulRM10 It's been updated, lmk if you see anything to be changed in this association.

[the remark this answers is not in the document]
saulRM10: I checked into your branch but i get issues while running the new migrations. The last one is failing :
image

I see you used `BIGINT` , and Roles model has that too but in your migration for roles you did not include that change so the database still a `Integer` type. Not sure for roles and permissions we would need a `BIGINT` for ids.

saulRM10 (changes requested):
I am getting an issue when running the last migration for `RolePermissions` 

https://github.com/cascarita-io/cascarita/pull/167#discussion_r1942183137

server/migrations/20250203203552-create-role-permissions-table.js
```diff
@@ -40,14 +30,42 @@ module.exports = {
       },
     });
 
+    await queryInterface.addConstraint("RolePermissions", {
```
saulRM10, line 33: For these fk, it should be fine to keep them in the `permission_id: {..` and  `role_id: {..` , since the table is dropped (in the order the down of migrations will get ran) The only issue is when the constraint is dropped while there are existing fk references. The main issue is with the drop() of the constraigt migrations

server/migrations/20250203202920-add-unique-constraint-to-userroles.js
```diff
@@ -10,7 +10,10 @@ module.exports = {
     });
   },
 
-  async down (queryInterface, Sequelize) {
+  async down(queryInterface, Sequelize) {
+    await queryInterface.removeConstraint("UserRoles", "UserRoles_ibfk_1");
+    await queryInterface.removeConstraint("UserRoles", "UserRoles_ibfk_2");
+
```
saulRM10, line 16: This is good, you remove the fk and then the constraint but we you need to add back the fk constraint or else the User and Role wont be linked to the UserRoles table.

saulRM10 (changes requested):
Need to fix constraint migration, permissions migration should be reverted since no issues are created when just dropping the table. Sent you a dm for changes I suggest for  the constraint migration

server/migrations/20250203202920-add-unique-constraint-to-userroles.js
garcialuis: With the last update the foreign keys will be reestablished as part of the down migration.

saulRM10 (approved):
Looks good, Thanks Luis for this pr

PR #182, opened 2025-02-14
Fix: Restore deleted account routes

saulRM10:
### Description

These routes got deleted during a refactor, needed for stripe process

### Checklist

### DIFF
--- a/server/routes/account.routes.js
+++ b/server/routes/account.routes.js
@@ -16,5 +16,8 @@ module.exports = (checkJwt) => {
     AccountController.getClientSecret,
   );
 
+  router.get("/:group_id", AccountController.getAllAccountsByGroupId);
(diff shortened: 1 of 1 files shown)


server/routes/account.routes.js
```diff
@@ -16,5 +16,8 @@ module.exports = (checkJwt) => {
     AccountController.getClientSecret,
   );
 
+  router.get("/:group_id", AccountController.getAllAccountsByGroupId);
+  router.get("/key/publishable", AccountController.getPublishableKey);
```
garcialuis, line 20: If the publishable key is not user facing then it is better to not expose an endpoint for it.

garcialuis (changes requested):

garcialuis (approved):

PR #150, opened 2024-11-08
fix: avoiding unnecessary pop up for register modal

abanuelo:
### Description

Issue is addressing #140 as even after you registered, upon refreshing the home screen you would get a flash of the registration modal, that no longer happens see gif below for an illustrated example:
Recording 2024-11-07 at 22 07 31

### Issue Link

#140  

### Checklist

- [ ] I have tested the changes locally by running `npm run test`
- [ ] I have added or updated relevant documentation
- [ ] I have autoformatted the code by running `npm run eslint`
- [ ] I have added test cases (if applicable)

PR #183, opened 2025-02-14
fix: linearize Home experience

abanuelo:
### Description

Introduces:
- Linear home experience where the tables give an indication as to what season, league, and/or division an entity is tied to.
- In the `POST` create modals, made sure to include extra information about the entity types in order to link the created item with its parent-level model
- Adding `404` page for frontend routes
- In order to prevent multiple calls to auth0 to get a token to then be able to get the current logged in user to get their `group_id` in order to pass that into several endpoints needed for the linearized experience, made it convenient to store the `group_id` as a Cookie to prevent un necessary queries to Auth0
- Adding fixes from #150 into this PR to create a crisp home experience where the registration league modal only appears where necessary
Recording 2025-02-14 at 11 08 33

### Issue Link

https://cascarita.atlassian.net/browse/CV-25

### Checklist

### Additional Notes

Note there are some `TODO` comments left in this PR but will be deferred to future PRs in order to prevent this single PR from getting to large. More notably, changes to `PATCH` existing models (ie Leagues, Teams, Seasons, etc), need to be cleaned up in order to work properly.


garcialuis (approved):
Changes LGTM, thank you for adding the teams functionality as well. I tested adding teams and it's working as expected now.

PR #184, opened 2025-02-16
User and Form table Updates and new UserSessions Table

JesusGo115:
### Description

Updated the Users table to have "address", "phone_number", "date_of_birth", and "internally_created" columns
Updated Forms table to have "form_type" column
Created new table "UserSessions" to connect Users to Forms for payments when seasons or divisions end/change

### Issue Link

N/A

### Checklist


server/migrations/20250216220249-update-user-table-with-address-dob-phone-num.js
```diff
@@ -0,0 +1,36 @@
...
+      allowNull: true,
+    });
+
+    await queryInterface.addColumn("Users", "date_of_birth", {
+      type: Sequelize.DATEONLY,
+      allowNull: true,
+    });
+
+    await queryInterface.addColumn("Users", "phone_number", {
+      type: Sequelize.STRING(10),
```
garcialuis, line 17: To not lock ourselves to numbers in the U.S. we should play it safe and have this be STRING(20) instead: The longest possible phone number, according to the E.164 standard, can have a maximum of 15 digits.

server/models/user.js
```diff
@@ -98,6 +98,28 @@ module.exports = (sequelize, DataTypes) => {
...
+        allowNull: true,
+      },
+      phone_number: {
+        type: DataTypes.STRING(10),
+        allowNull: true,
+        validate: {
+          len: {
+            args: [9, 10],
+            msg: "phone number must be between 9 and 10 characters",
+          },
```
garcialuis, line 116: Update the string type as well. IMO we can leave the validation as is since we are only expecting U.S. numbers at the moment.

garcialuis (changes requested):

server/migrations/20250216220249-update-user-table-with-address-dob-phone-num.js
JesusGo115: Donzo

saulRM10 (approved):
Thanks chuy for making all the changes !

garcialuis (approved):
LGTM, thanks Chuy!

PR #189, opened 2025-02-17
Fix/Stripe Payment Component Not Rendering

saulRM10:
### Description

 **Stripe form payment block broken**
- fixed it by using @abanuelo cookie group id solution

**Creating forms is broken  !**
-  FormPayment Table added but column used in the `Forms` table was name 'form_type' on the migration. But models were calling 'form_type_id' so i just made sure the models referencing form type were renamed to `form_type` 
- Since a `form_type` is required, and there is only one default on the database.  I set it as the default value on the controller function where the form is created

### Testing

1. Go to docker desktop -> click cascarita continer -> open the server container -> go to the `Exec`tab -> run seeder file: `npx sequelize-cli db:seed --seed 20240902193649-demo-stripe-user.js`
3. Creat a form with a payment block, drafggin the payment block should not break form builder
4.  Preview form, should see the Stripe Payment Component rendered!
5.  Returning back your db to normal since db changes are branch agnostic, if you checked into this branch changes where done to your database, here is how you can undo them to return to the state before you checked into the branch:
      a. Undo seeder file in `Exec` tab run : `npx sequelize-cli db:seed:undo --seed 20240902193649-demo-stripe-user.js`
 

### Checklist

### DIFF
--- a/client/src/components/DragAndDropComponents/DraggablePayment/DraggablePayment.tsx
+++ b/client/src/components/DragAndDropComponents/DraggablePayment/DraggablePayment.tsx
@@ -13,6 +13,9 @@ import nullthrows from "nullthrows";
 import { useAuth0 } from "@auth0/auth0-react";
 import { DraggableProps } from "../types";
 import { useQuery } from "@tanstack/react-query";
+import Cookies from "js-cookie";
+import { fetchUser } from "../../../api/users/service";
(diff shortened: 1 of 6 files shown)


abanuelo:
@saulRM10 added a state variable to prevent creation of multiple payment intents within the form

abanuelo (approved):
LGTM!

PR #192, opened 2025-02-20
Feature: Handle Stripe Payments

saulRM10:
### Description

This does a couple of things but over all when a form has a payment block and user fills it out we now keep track of it on our database from when they open the stripe payment element to when they entered their card info and submitted the form ! :

1. webhooks for payment intent events , at the moment only handles when a payment intent is submitted (aka: user provides card info)
2. `formpayment.controller.js `file created with functions to : 
- `createStripeFormPayment()` :  creates `FromPayment` record for a form that has a stripe payment element is rendered
- `connectResponseToFormPayment()` : a function to link `FromPayment` record  to a `Response` record from Mongo  collection
- `updateStripePayment()` : a function that gets used by the webhook to update the status of the stripe payment info, for now its when a user fills out the stripe payment element  
- `findFormPaymentByPaymentIntentId()` : a function to look up `FromPayment` record via a `payment_intent_id` 

3. migration to add a new column to the `FormPayments` table to track `Response` (mongo collections) response document ids as strings (for example `123abc` and not the `ObjectId('123abc'`) 
4. front end code to pass back payment intent id*

### NOTE: 
**I need help with the react typescript code** So i need the `payment_intent_id` when a payment is completed on the `FormPage.tsx` , here was my attempt to get the data to request body, could use feedback and refactor help of my solution

commit hash: 01d1eff86adca7fc41d6d23a0e4cf8935fd90032 

### Testing 

- I will have to hop on a call to show you all how to test!

### Issue Link

[Jira Card: CV-75 ](https://cascarita.atlassian.net/browse/CV-75)

### Checklist

### DIFF
--- a/client/src/api/forms/service.ts
+++ b/client/src/api/forms/service.ts
@@ -41,7 +41,7 @@ export const updateForm = async (
   formId: string,
   title: string,
   description: string,
-  user: User | null,
+  user: User | null
 ) => {
   const formData = {
     form_data: {
@@ -108,7 +108,7 @@ export const createMongoForm = async (
   title: string,
   description: string,
   groupId: number | undefined,
-  userId: number | undefined,
+  userId: number | undefined
 ) => {
   const formData = {
     title,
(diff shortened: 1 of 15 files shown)


abanuelo (approved):
Approving on my end @garcialuis and I worked through with Saul to get FE more uniform

garcialuis (approved):
LGTM!

PR #194, opened 2025-02-21
feat: send form via email template

abanuelo:
### Description

Configuring Amazon SES service with template to be able to send out cascarita forms via email. P**lease note that only my personal email address has been used as a verified receiver as we wait to get production SES account access (currently only in sandbox meaning all receivers need to be verified in order to receive an email)**

See existing demo below:
Recording 2025-02-20 at 16 21 18

### Checklist


garcialuis (approved):
LGTM

PR #199, opened 2025-02-23
Capture Payment Intent Endpoint and Switch Case

JesusGo115:
### Description

1. user toggles approve :
2. `POST` request to `api/forms/status`
3. route hits `formPayment.controller.js @ updatePaymentStatus()`
     - calls `AccountController.capturePaymentInt( );` 
     -  This function now checks if payment intent has been captured (ran into an error when capturing an already captured payment intent) via stripe api call
     -  captures payment intent via stripe api call
     -  store formatted answers and user selected status on the Response document (updated the Response Mongo "Schema" ) 
     - updates the `FormPayment` entry with the admin's id and when the admin toggled the approval selection 

4.  WEB HOOK : Once there is an update/change to a `payment_intent` stripe will sent a HTTP Request to `/api/webhook/stripe/payment` end point on our server
     - that route hits: `formTransaction.controller.js @ handleEvent() `where we handle 2 cases: 
     - When an api call for capturing a payment intent is used then the `event.type` will be a type of : [`payment_intent.succeeded`](https://docs.stripe.com/api/events/types#event_types-payment_intent.succeeded) and we get a payment intent object in the request body
     - We handle that event, we do 3 process; 
     -  1)` formPayment.controller.js @ updateStripePayment` is called to update the status of the `FormPayment` entry and the corresponding Response document with the status along with other payment data for the front end.  We can look up a `FormPayment` entry by the payment intent id, and then we can get the response document id from that found `FormPayment` entry  up using the payment intent id 
     - 2) `formPaymentController.handleUserUpdateStrip() ` is called to get all the data needed for the creation of resources. It does the same type of queries; look up entry in `FormPayment` by using the payment intent id , then using the found `FormPayment` entry's `response_document_id` we make a query to get the `formatted_answers` object in the document. 
     - 2a) Once all that data is retrived, check the status. If `approved` then it proceeds to call `UserController.createUserViaFromResponse()`
     -  3)  `UserController.createUserViaFromResponse()` This function is used to create corresponsding resources; 
     - Finds or creates a user in the database based on their email and group ID
     - If user already exists: updates their profile information (name, photo, contact details, etc.)
     - If user doesn't exist: creates a new user record with the provided information
     - Marks all users as "internally_created" with this flag set to true
     - Creates or retrieves a session for the user through SessionController
     - Associates the user with their session and team
     - Assigns the "Player" role to the user
     - Uses a database `transaction` to ensure all operations succeed or fail together

6. Error Handling:
     - returning an object , if no error: { `success`: boolean , `data`: depends-on-function , `status`: 200/201 } 
     - if error, object changes a bit  : { `success`: boolean , `error`: depends-on-function , `status`: 404/500 but depends on what caused the error} 

NOTE: With the web hook we always return a `200` to stripe because we got the request. If we have server error thats on us and we should handle them by shooting error messages but not by returning **_that_** status code to stripe.

### Checklist


server/controllers/account.controller.js
```diff
@@ -232,6 +234,134 @@ const AccountController = function () {
...
+
+      if (retrievePaymentIntent.data.status === "succeeded") {
+        return {
+          success: true,
+          data: retrievePaymentIntent.data,
+          status: 200,
+        };
+      }
+
+      const paymentIntent = await Stripe.paymentIntents.capture(
```
garcialuis, line 275: What are the possible values to get here? We should also validate `paymentIntent` before moving forward.

server/controllers/account.controller.js
```diff
@@ -10,7 +12,7 @@ const {
   Group,
   FormPayment,
 } = require("../models");
-const FormPaymentController = require("./formPayment.controller");
+//const FormPaymentController = require("./formPayment.controller");
```
garcialuis, line 15: We can get rid of this if no longer used.

garcialuis (changes requested):

server/controllers/account.controller.js
saulRM10: This returns a `payment intent` object,

garcialuis (approved):
Approved, but before merging let's create a story for us to implement our own retry mechanism for requests received by the webhook.

garcialuis:
Story: [CV-94](https://cascarita.atlassian.net/browse/CV-94?atlOrigin=eyJpIjoiM2ViOTQ3MWJhMzMwNDZkN2I5YWFiMTk0NTFlZjc0ZTQiLCJwIjoiaiJ9)

server/controllers/formPayment.controller.js
```diff
@@ -2,61 +2,106 @@
...
+  var connectResponseToFormPayment = async function (responseData, responseId) {
     const paymentEntry = responseData.find(
       (item) => item.field?.type === "payment" && item.paymentIntentId,
     );
 
     if (!paymentEntry) {
-      return;
+      return {
+        success: false,
+        error: "no payment entry forund in response data",
```
garcialuis, line 19: Typo

garcialuis (changes requested):

garcialuis (changes requested):
Cash payment responses also need to be handled. Currently when there is a cash payment response there is no payment intent data as part of the `responseData` that gets passed into the `FormPaymentController.connectResponseToFormPayment(...)`.

server/controllers/formPayment.controller.js
saulRM10: fixed

saulRM10:
> Cash payment responses also need to be handled. Currently when there is a cash payment response there is no payment intent data as part of the `responseData` that gets passed into the `FormPaymentController.connectResponseToFormPayment(...)`.

- latest commit fixes the issue

garcialuis (approved):

PR #201, opened 2025-02-23
Feat: Unique group codes

garcialuis:
### Description
Implements unique group codes. These codes are now required and checked by the registration api to ensure that only users with the correct code can join an existing group.

### Issue Link
[CV-83](https://cascarita.atlassian.net/browse/CV-83?atlOrigin=eyJpIjoiOTliZDc0Y2ZiNjRmNGFiM2IzNmFkNTkyZmI5ZWI1YjMiLCJwIjoiaiJ9)

### Checklist

- [x] I have tested the changes locally

### Additional Notes
Screenshot 2025-02-22 at 11 28 51 PM

### DIFF
--- a/client/src/api/users/types.ts
+++ b/client/src/api/users/types.ts
@@ -32,6 +32,7 @@ export interface RegisterUser {
   state: string | null;
   zipCode: string | null;
   logoUrl: string | null; // still need to set up s3 buckets for images so for now we wont collect this and just set it to null
+  group_code: string | null;
   token: string | null;
 }
 
--- a/client/src/components/RegistrationModal/RegistrationModal.tsx
+++ b/client/src/components/RegistrationModal/RegistrationModal.tsx
@@ -28,6 +28,7 @@ const initialState = {
   city: "",
   state: "",
   zipCode: "",
+  group_code: "",
 };
 
 function reducer(state: State, action: Action) {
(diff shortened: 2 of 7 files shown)


abanuelo (changes requested):
Couple of bugs:

1. Since `group_code` is required by the form, when you try to fill out the registration for a new org, it will block the submission
2. When I submit a `group_code` that does not belong to the group I have selected the modal exits but we really want to display those errors in the modal preventing the user form creating an account
3. Maybe make an API controller method that given a groupId can find the group_code and verify if what the user entered match or not, can help point 2. above

Let me know if you need any assistance there @garcialuis good work!

garcialuis:
> Couple of bugs:
> 
> 1. Since `group_code` is required by the form, when you try to fill out the registration for a new org, it will block the submission
> 2. When I submit a `group_code` that does not belong to the group I have selected the modal exits but we really want to display those errors in the modal preventing the user form creating an account
> 3. Maybe make an API controller method that given a groupId can find the group_code and verify if what the user entered match or not, can help point 2. above
> 
> Let me know if you need any assistance there @garcialuis good work!

@abanuelo 3 is already happening and the controller is returning an error [here](https://github.com/cascarita-io/cascarita/pull/201/files#diff-03c5cfdce3d326081ea81ea71736a28085c612f08e4e69a1fde34f1a7519c2bdR96-R97). We just need to display the error in the modal and **not close the modal on submission** when there are errors.

abanuelo:
Added UI to show group code registration errors
Recording 2025-02-23 at 21 35 14

abanuelo (approved):

PR #203, opened 2025-02-23
fix: bug bash session pt.1

abanuelo:
### Description

When implementing some changes @JesusGo115 and @saulRM10 noticed some bugs this PR attempts to address some of those bugs namely:

1. Typing `Enter` on any of the blocks would close out the survey. I have made the key bindings linked to the next and submit buttons in the Form experience
2. Players page would show some data but then it would immediately display an error. Fixed that by adding check for player role in controller method
3. `make build` was failing due to some `eslint` errors; that has been fixed as needed
4. Allowing for users to select cash options form the stripe payment block

### Checklist

### DIFF
--- a/client/src/api/forms/service.ts
+++ b/client/src/api/forms/service.ts
@@ -240,12 +240,10 @@ export const sendEmail = async (formLink: string, email: string) => {
   }
 };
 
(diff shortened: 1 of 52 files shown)


saulRM10 (approved):
LGTM !

PR #208, opened 2025-02-25
fix: get group id upon first registration to show leagues

abanuelo:
### Description

When users first log in should be able to render Leagues for current group selected immediately:

Recording 2025-02-25 at 09 44 22

### Checklist


jipatino98 (approved):
Nice work! Also good idea on creating the context for the group Id.

PR #217, opened 2025-02-28
adds api endpoint to return data needed for user settings' page

garcialuis:
### Description

This PR is for the User's Settings page. It includes:
- API endpoint to that returns User and User's Group info
- Front end changes to display the Group code
- Fixes a bug where the NavBar loses its active styling when the user clicks away

### Issue Link
Part of [CV-60](https://cascarita.atlassian.net/browse/CV-60)

### Screenshots

https://github.com/user-attachments/assets/daad4086-d5c0-40e6-a522-bada8392eda0

### Additional Notes
- The group code will only be displayed for Staff or Admin Roles. (Subject to change in the future)
- Middleware placement was also updated to catch unhandled errors.
- If a user is not an Admin or Staff, then no Settings information is displayed. Should we populate info from auth0 if they are not admin?
image

### DIFF
--- a/server/controllers/user.controller.js
+++ b/server/controllers/user.controller.js
@@ -654,7 +654,6 @@ const UserController = function () {
         },
       });
 
-      console.log("UserRole", UserRole);
       if (!UserRole) {
         res.status(404).json({
           error: `no user role was found with user id ${user.id}`
--- a/server/controllers/user.controller.js
+++ b/server/controllers/user.controller.js
@@ -32,6 +32,7 @@ const {
   TeamsSession,
   Season,
   Team,
+  Group,
 } = require("../models");
 const Db = require("../models");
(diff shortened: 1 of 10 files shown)


server/controllers/user.controller.js
```diff
@@ -629,6 +630,67 @@ const UserController = function () {
...
+
+      if (!UserRole) {
+        res.status(404).json({
+          error: `no user role was found with user id ${user.id}`
+        });
+      }
+
+      let groupCode = group.group_code;
+
+      var userRole = await Role.findByPk(UserRole.role_id);
```
saulRM10, line 665: what if they have multiple roles ?

saulRM10 (changes requested):

abanuelo (approved):
LGTM! Changes look good!

PR #220, opened 2025-03-03
fix: accept/reject status button for specific indexes

abanuelo:
### Description

Bug introduced where selecting on changing the payment status always defaulted to the last row in the form responses table. Now the button can render the accept/rejection/cancel view for the specific row in question.
Recording 2025-03-02 at 19 21 44

### Checklist


jipatino98 (approved):
Nice work! No comments from me! Approved!

PR #53, opened 2024-05-11
feat!: Adding sentry logging and monitoring for our application

abanuelo:
##### Overview 

As part of being able to have robust logging and monitoring in place, I have configured our application with [Sentry](https://sentry.io/welcome/). This account is free for a single developer so we may need to share a single login screen and password as go forward to stay lean 💪 . Also as we expand the fee for this service is 26 dollars a month which is very reasonable given the number of features on this platform.

 #### Features!
- Get Performance insights such flame charts showing how fast our APIs are. This can help determine where our bottlenecks are at and to improve performance as we scale with the number of data requests we receive. 
- Get Error Insights that give a replay of exactly how users experience errors on the frontend
- Integrations with Jira and GitHub to file tickets and issues, respectively, when errors do assign to get people resolving the issues quickly. 
- Also as part of the integration with GitHub, it looks at the entire commit history and can identify the most likely commit that introduced the error
- Easily query performance insights based on development environment
- Generative AI features that can help us figure out the errors in real time

#### Demo!
In order to show the demo, I add the "Add League" button call a nonexistent function within our React code that generated the error. Below are the ways we can get insights into a specific error.  

https://github.com/cascaritaco/cascarita/assets/32311654/62941bd7-ec50-442a-95a9-312ef52cb362

#### Important Notes
In order for Sentry to be able to capture these metrics from our application its important to define the `SENTRY_AUTH_TOKEN` within the `.env` file. This secret has been stored on `AWS Secrets Manager` and I have defined a `loadSecrets.js` script that will pull the credentials and write to the `.env` file. In order to have that code work, you will need to define an `ACCESS_KEY` and a `SECRET_ACCESS_KEY` in your `~/.aws/credentials` directory. To learn more on how to create your own keys I will leave this Jira here for reference: https://cascarita.atlassian.net/l/cp/MwpAAG84

### Issue Link

N/A

### Checklist

- [X] I have tested the changes locally by running `npm run test`
- [X] I have added or updated relevant documentation
- [X] I have autoformatted the code by running `npm run eslint`
- [ ] I have added test cases (if applicable)

PR #222, opened 2025-03-03
feat: sentry logging

abanuelo:
## DO NOT MERGE YET UNTIL WE FINALIZE PRODUCTION

Closing out #53 as this replicates logging but adds it to FE and BE. 

Screenshot 2025-03-02 at 9 26 05 PM

### Issue Link

https://cascarita.atlassian.net/browse/CV-115

### Checklist


client/src/index.tsx
```diff
@@ -4,6 +4,22 @@ import { createRoot } from "react-dom/client";
 import { Auth0Provider, CacheLocation } from "@auth0/auth0-react";
 import { createBrowserHistory } from "history";
 import "./index.module.css";
+import * as Sentry from "@sentry/react";
+
+Sentry.init({
+  dsn: "https://c3264debab8e92379efef056a7005b80@o4508911694315520.ingest.us.sentry.io/4508911696609280",
```
garcialuis, line 10: Is this unique to us? Does it need to be changed and placed in an env var instead?

server/Dockerfile
```diff
@@ -2,15 +2,14 @@ FROM node:23-alpine
 
 WORKDIR /app
 
-# Copy package.json and package-lock.json from the root
-COPY ./server/package.json ./server/package-lock.json ./
+# Copy the rest of the application files
+COPY ./server/ .
+
+RUN rm -rf node_modules package-lock.json
```
garcialuis, line 8: I think these changes are not relevant anymore since it's been updated in a recent PR.

server/app.js
```diff
@@ -44,8 +46,8 @@ const checkJwt = require("./checkJwt");
 app.use(bodyParser.json());
 app.use(bodyParser.urlencoded({ extended: true }));
 
-// Error handler should be the last middleware used
-app.use(Middlewares.errorHandler);
+// // Error handler should be the last middleware used
+// app.use(Middlewares.errorHandler);
```
garcialuis, line 50: We can leave this commented until we update the client to handle the error responses but I suggest moving them down to line 84 to avoid re-enabling the middleware before Sentry in the future.

server/instrument.js
```diff
@@ -0,0 +1,17 @@
+// Import with `import * as Sentry from "@sentry/node"` if you are using ESM
+const Sentry = require("@sentry/node");
+
+Sentry.init({
+  dsn: "https://b2949253b19698b31aaaee4c49d722f1@o4508911694315520.ingest.us.sentry.io/4508911777742848",
```
garcialuis, line 5: Same here as before. Env var?

client/src/index.tsx
garcialuis: Also how would we differentiate between staging and production? Or dev and production

garcialuis (changes requested):
Thanks for adding this Armando, it'll be extremely useful for us to monitor. 

Marking as request for changes to prevent merge before comments are discussed.

server/Dockerfile
abanuelo: Correct let me go ahead and remove them from this PR good catch!

server/app.js
abanuelo: Sure works for me! Will take this on as well!

server/instrument.js
abanuelo: See comment above.

client/src/index.tsx
garcialuis: Sounds good, that was my main concern but it is addressed in that docs page. Thanks for including those

garcialuis (approved):

PR #225, opened 2025-03-04
Fix/Allow admins to Approve payments

saulRM10:
### Description

Before only the owner of the stripe account could Approve a payment in teh response tab without returning a server error. 

1. updated the `formPayments` table to refrence the `userStripeAccounts` table `stripe_account_id` over the `user_id`

### Checklist

### DIFF
--- a/server/controllers/formPayment.controller.js
+++ b/server/controllers/formPayment.controller.js
@@ -199,7 +199,7 @@ const FormPaymentController = function () {
     }
   };
 
-  var updatePaymentStatus = async function (req, res, next) {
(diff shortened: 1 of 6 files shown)


server/controllers/formPayment.controller.js
```diff
@@ -199,7 +199,7 @@ const FormPaymentController = function () {
     }
   };
 
-  var updatePaymentStatus = async function (req, res, next) {
+  var updatePaymentStatus = async function (req, res) {
```
garcialuis, line 202: What's the reasoning behind removing `next` here?

garcialuis (approved):

PR #226, opened 2025-03-04
Fix: Increase client max body size

garcialuis:
### Description

Increases the client max body size. 
The default max by nginx is 1mb. We are accepting user images that are up to 8mb large but nginx was only allowing requests that were up to 1mb. So any requests with images that exceeded the default value were being rejected by nginx.

Here the max size is set to 15mb to allow room for larger requests while still setting a reasonable limit.
### Issue Link

[CV-103](https://cascarita.atlassian.net/browse/CV-103?atlOrigin=eyJpIjoiMTQxNmEzZmNiYTYyNDZlMGI2ZTQ4M2Y5YzkyODg5MGUiLCJwIjoiaiJ9)

### DIFF
--- a/client/src/components/FileUpload/FileUpload.tsx
+++ b/client/src/components/FileUpload/FileUpload.tsx
@@ -62,7 +62,7 @@ const FileUpload: React.FC<FileUploadProps> = ({ className, setFileValue }) => {
     accept: {
       "image/*": [],
     },
-    maxSize: 1024 * 1000 * 8,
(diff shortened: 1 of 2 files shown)


abanuelo (approved):
LGTM! I had to also increase MB in FileUpload to 10MB but tested that it does work:

Screenshot 2025-03-04 at 7 32 14 PM

And here is the image I used for reference:
8mb

saulRM10:
checked out into the branch and tested the whole flow . good work

PR #227, opened 2025-03-05
fix: assign role on registration

villaleo:
### Description

When a user creates an account, they will be assigned the role of "Staff" with ID 1.

### Issue Link

[Jira Ticket](https://cascarita.atlassian.net/browse/CV-101?atlOrigin=eyJpIjoiNWNjNzg2M2Q2YTMxNDg3M2I1MzdhYmU2NWQzYzUyNjEiLCJwIjoiaiJ9)

### DIFF
--- a/server/controllers/user.controller.js
+++ b/server/controllers/user.controller.js
@@ -146,6 +146,12 @@ const UserController = function () {
       await User.build(newUser).validate();
       const result = await User.create(newUser);
 
+      // Assign the role of "Admin" to the new user.
+      await UserRoles.create({
+        user_id: result.id,
+        role_id: 1,
+      });
+
       return res.status(201).json(result);
     } catch (error) {
       console.error(error);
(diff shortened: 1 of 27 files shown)


server/controllers/user.controller.js
```diff
@@ -146,6 +146,12 @@ const UserController = function () {
       await User.build(newUser).validate();
       const result = await User.create(newUser);
 
+      // Assign the role of "Admin" to the new user.
+      await UserRoles.create({
+        user_id: result.id,
+        role_id: 1,
```
saulRM10, line 152: I recommend updating the role assignment to use const `adminRole = await Role.findOne({ where: { Name: 'Admin' } }) ` instead of hardcoding the `role_id` as 1. Not sure what role id Admin has on prod but on my local its not 1

saulRM10 (changes requested):
Looks good, left a comment

server/controllers/user.controller.js
villaleo: Thanks for your feedback! I totally forgot to query the db haha

saulRM10 (approved):
LGTM , great job Leo

villaleo:
Great, thank you!

PR #228, opened 2025-03-06
fix: Docker dev experience

abanuelo:
### Description

Fixing broken developer experience:
1. The `@hookform/resolvers` is added as a `devDependencies` which makes it so that these dependencies do not get added when we run `make build`. If the package needs to exist in prod please add it to `dependencies` section  in the `package.json`
cc: @jipatino98 

2. We were mounting our local node modules within the Docker containers for server and client. Adding checks to remove that directory prior to install dependencies. This eliminates the need for cross-platform install of the `sharp` dependency needed for image upload
cc: @garcialuis 

3. Fixing our [`@tanstack/react-query` dependency](https://www.npmjs.com/package/@tanstack/react-query) to a static version as the latest version introduces breaking changes with `useQuery` and `useQueries` hook to pass `client QueryClient` which will lead to a larger refactor. Will add another ticket to handle this later to our backlog
cc: @garcialuis @jipatino98 

### Issue Link

Not linked to Jira but part of dev experience

### Checklist


saulRM10 (commented):
Checked into the branch and all dependencies installed , `zod` ones did NOT need me to install it separately and manually!  But ill defer to @garcialuis  for the approval

garcialuis (approved):
LGTM!

PR #229, opened 2025-03-06
Feature: Admin can decline stripe payments

saulRM10:
### Description

Admin can toggle 'rejected' a payment from the Form Response ui:
1. now when the toggle is switched from `approve` or `rejected` then this function will handle a cancel of payment intent or the capturing of a payment intent (capture means to collect the funds) : `account.controller.js @ processPaymentIntent()`

### Checklist


server/controllers/account.controller.js
```diff
@@ -216,56 +216,31 @@ const AccountController = function () {
...
-      );
-
-      if (!paymentIntent) {
-        return {
-          success: false,
-          error: `no stripe payment intent found with id of: ${paymentIntentId}`,
-          status: 404,
-        };
-      }
+      await Stripe.paymentIntents.capture(paymentIntentData.id, {
```
garcialuis, line 238: Here we are no longer checking to make sure we are get an object back from stripe?

server/controllers/account.controller.js
```diff
@@ -216,56 +216,31 @@ const AccountController = function () {
...
       }
 
-      if (retrievePaymentIntent.data.status === "succeeded") {
+      const { userId, formPayment, paymentIntentData } = preparedInfo;
+
+      if (paymentIntentData.status === "succeeded") {
         return {
           success: false,
-          data: `payment intent of status succeeded can not be captured: ${retrievePaymentIntent.data}`,
+          data: `payment intent of status succeeded cannot be captured: ${paymentIntentData.id}`,
```
garcialuis, line 233: IMO I would change text to: "payment intent status is succeeded cannot be re-captured", but up to you.

server/controllers/account.controller.js
```diff
@@ -312,50 +346,104 @@ const AccountController = function () {
...
       };
     }
   };
 
+  var preparePaymentIntentOperation = async function (paymentIntentId, email) {
+    try {
+      const { id } = await User.findOne({
+        where: { email: email },
+      });
+
```
garcialuis, line 384: Is the email validated sometime before this function is called?

saulRM10: this email is from the user who toggled the UI switch

saulRM10: I can add a check , just to make sure we get an id

server/controllers/account.controller.js
saulRM10: If there is an error, it will go to the` catch()` a common error is capturing something with status of `succeeded` and i have a check before this line to make sure we don't capture if status is  `succeeded`

garcialuis (approved):
Thanks for getting those changes in, LGTM!

PR #242, opened 2025-03-10
Feat/add validation response pages

jipatino98:
### Description

This PR introduces a dedicated schema file that handles the validation rules of the various form blocks. Some rules that we created were:
- short text fields must be less than 30 characters
- liability fields will have a default value of `checked` and is required
- phone number must be at least 9 digits and no greater than 20 digits
This will give us a centralized file that will contain all the validations, which is able to easily be modified.

### Issue Link

[Jira Ticket](https://cascarita.atlassian.net/browse/CV-98)

### Checklist


abanuelo (approved):
LGTM!

PR #253, opened 2025-03-13
Feature: Filter out form payments that do not have a response id

saulRM10:
### Description

We can filter out formPayments with `response_document_id: null`. When a form is submitted, a `response_document_id` is linked to the corresponding `formPayment`. Therefore, in the Forms/Response view, we should only handle `formPayments` that have a valid `response_document_id`. Currently, the view uses every `formPayment` entry created for the form, regardless of submission status. Which causes responses to have missing status fields on the UI

### Checklist


garcialuis (approved):
Nice and simple, lgtm thank you!

PR #254, opened 2025-03-13
fix: enabling Sentry only for prod

abanuelo:
### Description

[As noted in Sentry docs](https://docs.sentry.io/platforms/javascript/configuration/options/#enabled), conditionally rendering the `Sentry.init` call in order to only show Sentry on production

### Checklist


saulRM10 (approved):
LGTM

PR #259, opened 2025-03-14
Feature: Store Stripe Payment Web-hook Events

saulRM10:
### Description

1.  migration created for a new table and model, `stripeEvents` . We will use this table to store webhook events. Store the `event_id`, `event_type`, `event_id` , `payload` , `internal_satatus` , `stripe_account_id` , `livemode` (these columns are also indexed on the db to it is easy to query and debug in the future)
2. `stripeEvent.controller.js` file created with a function that handle checking for duplicates , ignoring certain events (at the moment this table is only going to store payment events, we have two webhooks but the other webhook will not receive no where near events as the payment one. The other webhook is the `connectedAccount.controller.js` which handles updates to our users stripe account) and handle retries since stripe will use the same `event.id` (we store that) 
3. updated the way we handle webhook payments, by breaking it down to 2 functions. The original `handleEvent()` that stripe wehook will send the request too (sends it to a route in our server but that route calls this function). This function verifies the stripe request (make sure its a real stripe event) and **RETURNS 200** right away (stripe docs recommend this, stripe wants to know that our server got the event. What happens after that is on our server to handle and NOT SEND BACK A STATUS CODE. Then the function calls a `processEvent()` where we do all the db/bussiness logic. I make it so it never returns a status, only console logs for easier debugging and since we are now storing the event on the `stripeEvents` table, if something goes wrong where there was a db/logical error we can manually fix it since we have all the relevant event data on the db. @abanuelo  maybe we connect it with sentry, not sure if i need to throw an error to trigger sentry ?

### Checklist

### DIFF
--- a/server/controllers/stripeEvent.controller.js
+++ b/server/controllers/stripeEvent.controller.js
@@ -13,6 +13,7 @@ const StripeEventController = function () {
       if (ignore) {
         return {
           success: true,
+          skip: true,
           data: `not tracking events of type : ${eventType}`,
           status: 200,
         };
@@ -24,7 +25,10 @@ const StripeEventController = function () {
 
       if (existingEvent) {
         return {
-          success: false,
+          success: true,
+          skip:
+            existingEvent.status === "completed" ||
+            existingEvent.status === "processing",
           data: `event ${eventId} already processed, skipping`,
           status: 200,
         };
@@ -40,6 +44,7 @@ const StripeEventController = function () {
 
       return {
         success: true,
+        skip: false,
         data: `successfully saved the event: ${eventId}`,
         status: 200,
       };
--- a/server/controllers/stripeEvent.controller.js
+++ b/server/controllers/stripeEvent.controller.js
@@ -65,7 +65,12 @@ const StripeEventController = function () {
         { where: { event_id: eventId } },
       );
     } catch (error) {
-      console.error(`Failed to update event status: ${error.message}`);
+      console.error({
+        event: "event_internal_status_update_failed",
+        eventId: eventId,
+        internal_status: status,
+        error: error.message,
+      });
     }
   };
 
--- a/server/controllers/stripeEvent.controller.js
+++ b/server/controllers/stripeEvent.controller.js
@@ -57,6 +57,17 @@ const StripeEventController = function () {
     }
   };
 
+  var updateEventStatus = async function (eventId, status) {
+    try {
+      await StripeEvent.update(
+        { status: status },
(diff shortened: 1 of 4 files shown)


abanuelo (approved):

server/controllers/stripeEvent.controller.js
```diff
@@ -0,0 +1,92 @@
+"use strict";
+
+require("dotenv").config();
+
+const { StripeEvent } = require("../models");
+
+const StripeEventController = function () {
```
garcialuis, line 7: Since this is not receiving any requests directly from the client I would have this as a service rather than a controller.

server/controllers/stripeEvent.controller.js
```diff
@@ -0,0 +1,92 @@
...
+        data: `successfully saved the event: ${eventId}`,
+        status: 200,
+      };
+    } catch (error) {
+      return {
+        success: false,
+        error: `error handling storing Stripe event on databsae : ${error.message}`,
+        status: 500,
+      };
+    }
```
garcialuis, line 58: It would be helpful to log any errors here, specially on the creation of the StripeEvent. Without that entry we won't have a record of the event. A log can help us identify issues in sentry.

server/controllers/stripeEvent.controller.js
```diff
@@ -0,0 +1,92 @@
...
+      await StripeEvent.update(
+        { internal_status: status },
+        { where: { event_id: eventId } },
+      );
+    } catch (error) {
+      console.error({
+        event: "event_internal_status_update_failed",
+        eventId: eventId,
+        internal_status: status,
+        error: error.message,
```
garcialuis, line 72: Do we want to display this error message in the console?

server/controllers/stripeResources/formTransaction.controller.js
```diff
@@ -2,106 +2,174 @@
 require("dotenv").config();
 const Stripe = require("stripe")(process.env.STRIPE_TEST_API_KEY);
 const FormPaymentController = require("../formPayment.controller");
+const StripeEventController = require("../stripeEvent.controller");
 
 const FormTransactionController = function () {
-  const endpointSecret = process.env.STRIPE_CONNECTED_ACCOUNTS_WEBHOOK_SECRET;
+  const endpointSecret = process.env.STRIPE_PAYMENTS_WEBHOOK_SECRET;
```
garcialuis, line 8: Let's make sure secret manager is updated with the new var name.

PR #261, opened 2025-03-15
feat: add expiry status form responses 

abanuelo:
### Description

Adding an expiry status to show expired payment captures in the form responses view. Also updating the share modal for forms since it had some bugs

Screenshot 2025-03-14 at 9 27 28 PM

Screenshot 2025-03-14 at 9 24 54 PM

Screenshot 2025-03-14 at 9 01 58 PM

### Checklist

### DIFF
--- a/client/src/components/FormResponses/FormResponses.tsx
+++ b/client/src/components/FormResponses/FormResponses.tsx
@@ -118,6 +118,12 @@ const FormResponses = ({ formId }: FormResponsesProps) => {
     return date.toLocaleDateString(undefined, options);
   };
 
+  const getExpiryDate = (dateString: string): Date => {
+    const date = new Date(dateString);
+    date.setDate(date.getDate() + 3);
+    return date;
+  };
+
   const formattedCurrency = formatCurrency(amount);
 
   useEffect(() => {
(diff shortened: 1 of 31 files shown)


client/src/components/FormResponses/FormResponses.tsx
```diff
@@ -118,6 +118,12 @@ const FormResponses = ({ formId }: FormResponsesProps) => {
     return date.toLocaleDateString(undefined, options);
   };
 
+  const getExpiryDate = (dateString: string): Date => {
+    const date = new Date(dateString);
+    date.setDate(date.getDate() + 3);
+    return date;
+  };
+
```
eli-gather, line 126: Anything in a component will rerender, so my advice is to have this helper outside of a component. I also don't know what this constant `+3` means can we have that constant outside of this component and call it `const DAYS_BEFORE_EXPIRY= 3`.

client/src/components/FormResponses/FormResponses.tsx
```diff
@@ -317,51 +323,60 @@ const FormResponses = ({ formId }: FormResponsesProps) => {
               <td>{paymentType[index]}</td>
               {formType === 1 && (
                 <td>
-                  <DropdownMenuButton
-                    className={styles.dropdown}
-                    trigger={StatusButton(status[index])}
-                  >
-                    <DropdownMenuButton.Item
-                      onClick={() => handleOpenPaymentModal(index)}
+                  {new Date() > getExpiryDate(submittedAt[index]) ? (
```
eli-gather, line 326: we need to figure out a way to make `date` a state. If I leave my computer on 5 min before the date expired and 10 min later, then technically I can still see this information / submit data after the expired date

client/src/components/FormResponses/FormResponses.tsx
```diff
@@ -118,6 +118,12 @@ const FormResponses = ({ formId }: FormResponsesProps) => {
     return date.toLocaleDateString(undefined, options);
   };
```
jipatino98, line 120: Let's start moving our helper functions into their own dedicated files. I created one in the same directory called `helper.ts`. This will help minimize the need to scan a lot of logic that we can easily abstract.

client/src/components/FormResponses/FormResponses.tsx
```diff
@@ -390,12 +405,18 @@ const FormResponses = ({ formId }: FormResponsesProps) => {
               <td>{email[index]}</td>
               <td>{`$${formattedCurrency[index]}`}</td>
               <td>
-                {paymentType[index] === "Credit Card / Stripe"
-                  ? getStatusOfStripePayment(
+                {paymentType[index] === "Credit Card / Stripe" ? (
```
eli-gather, line 408: I really wish `"Credit Card / Stripe"` was a payment type that we can share 🙏 

overall a good structure that I would have liked to see is ...
```suggestion
                {payment.type === "Credit Card" ? (
```

jipatino98 (approved):
Looks good! Just address the nit pick i left in `FormResponses.tsx`

client/src/components/StatusLabel/StatusLabel.tsx
```diff
@@ -54,13 +57,17 @@ const statusLabelStyling = (status: string) => {
         ? "#e9ffe8"
         : label === "rejected"
           ? "#ffeeee"
-          : "#dbe7f98f",
+          : label === "expired"
+            ? "#fff5e6"
```
eli-gather, line 61: Not in this PR but it'd be nice if we can have color tokens. So it'd look something like ..
```suggestion
            ? colorTokens.expired // "#fff5e6"
```

eli-gather (commented):
Overall looks good! Some immediate fixes, and others probably need some redesign on how we digest the data

client/src/components/FormResponses/FormResponses.tsx
abanuelo: Great suggestion I think for now we can defer this to another PR

client/src/components/StatusLabel/StatusLabel.tsx
abanuelo: Seems reasonable let me implement this!

client/src/components/FormResponses/FormResponses.tsx
abanuelo: Great point here. Let me think about this

client/src/components/FormResponses/FormResponses.tsx
abanuelo: Good eye @jipatino98 let's do it!

client/src/components/FormResponses/FormResponses.tsx
eli-gather: fwiw this is where having a global state for Date would be nice to have. Maybe we can pair on how to use redux to manage that 😄

jipatino98: Yeah let me know once someone works on it! I would like to see how to handle this. Would be able just to create a global state variable thru context instead or is it better with redux?

client/src/components/FormResponses/FormResponses.tsx
jipatino98: Good point! I was thinking that after V1, we can go back and put all our helper functions in a dedicated directory. That way we dont have to keep rewriting the same logic. The more specific ones like `getExpiryDate` can go within the same directory that they are used.

client/src/components/FormResponses/FormResponses.tsx
eli-gather: Yeah i think `useContext` will be better since it's not a complex state that we need to manage

PR #266, opened 2025-03-16
Fix: keep original image orientation

garcialuis:
### Description
This change is to fix image orientation for images included in registration forms.

Uses image's exif metadata to read the image orientation and include that in the final image that is uploaded to S3. 

### Issue Link
[CV-157](https://cascarita.atlassian.net/browse/CV-157?atlOrigin=eyJpIjoiMWZhMTNjMTlmMGZmNDVjZWIxMmJkMjk0ZDU3YjUxY2UiLCJwIjoiaiJ9)

### Additional Notes
Testing our latest release I noticed that some images were being rotated when they were uploaded through a phone. This PR is a fix for that issue.

### DIFF
--- a/server/controllers/s3.controller.js
+++ b/server/controllers/s3.controller.js
@@ -28,7 +28,10 @@ const uploadImage = async (req, res) => {
   }
 
   var metadata = await sharp(req.file.buffer).metadata();
-  const exif = exifReader(metadata.exif);
+  var exif;
+  if (metadata.exif) {
+    exif = exifReader(metadata.exif);
+  }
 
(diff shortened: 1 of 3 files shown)


saulRM10 (approved):
LGTM, checked into branch and tested. Ran into some errors but Luis solved them. 

NOTE: We should track what images on s3 are actually being used and where. Also we should be able to delete images when we delete team or a user (Future idea). Another issue is if a player uploads and image and goes back and forth on the photo question then the photo url is not included in the response. One thing to think about is if we can upload a photo to s3 when a user SUBMITS the full form.

PR #272, opened 2025-03-19
fix: properly displaying delete form error messages

abanuelo:
### Description

BE returns an `{'error': '...'}` response when a form is flagged as unable to be deleted as it contains form responses. Allowing FE to capture these changes

#### Failure to delete form with responses
Recording 2025-03-18 at 20 59 11

#### Successful form deletion
Recording 2025-03-18 at 20 59 55

### Issue Link

https://cascarita.atlassian.net/browse/CV-144

### Checklist


jipatino98 (approved):
Nice work! Tested on my machine and works as expected, approved!

PR #273, opened 2025-03-19
fix: properly showing expiry

abanuelo:
### Description

Expiry should only show if
- payment is of type `credit card/stripe`
- status is NOT `approved` or `rejected` 
- current date has exceeded the expiry limit

Screenshot 2025-03-18 at 9 48 26 PM

### Issue Link

https://cascarita.atlassian.net/browse/CV-164

### Checklist

### DIFF
--- a/client/src/components/FormResponses/FormResponses.tsx
+++ b/client/src/components/FormResponses/FormResponses.tsx
@@ -52,7 +52,7 @@ interface PaymentCaptureModalProps {
   handleStatusChange: (
     index: number,
     statusUpdate: "approved" | "rejected" | "pending",
-    response: Record<string, Answer>,
+    response: Record<string, Answer>
   ) => void;
 }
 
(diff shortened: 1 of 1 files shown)


client/src/components/FormResponses/FormResponses.tsx
```diff
@@ -384,16 +387,18 @@ const FormResponses = ({ formId, populateResponses }: FormResponsesProps) => {
...
                   ) : (
                     getStatusOfStripePayment(
                       status[index],
-                      formatDate(submittedAt[index], 3),
+                      formatDate(submittedAt[index], 3)
                     )
                   )
                 ) : (
-                  ""
+                  "N/A"
```
garcialuis, line 401: I like what you have here, we're now showing that it's expired. Should we also continue to show the date it expired?

garcialuis (commented):
Thanks for the fix @abanuelo , I just left one question regarding what we should show in the expiry column. Let me know what your thoughts are there. Overall looks good :)

client/src/components/FormResponses/FormResponses.tsx
abanuelo: Good question @garcialuis I think for now we can keep as is right?

garcialuis: Yeah, we can change based on feedback later.

garcialuis (approved):

PR #276, opened 2025-03-20
fix: bug bashes

abanuelo:
### Description

- fixed search bar across Users, Seasons, Divisions, and Teams pages
- broken `publish.yml` workflow
- only allow forms to be clickable if you are on desktop view in order to see form responses
- show delete league errors if league is linked to various seasons
Screenshot 2025-03-19 at 10 41 44 PM

- showing divisions by seasonId when creating team in `TeamForm` modal. Also displaying `{league_name} - {season_name}` convention when selecting season
Screenshot 2025-03-19 at 10 47 15 PM

- when editing divisions, they should only be allowed to edit name
Screenshot 2025-03-19 at 10 49 57 PM

### Checklist


server/controllers/league.controller.js
```diff
@@ -98,9 +98,22 @@ const LeagueController = function () {
...
     try {
+      const leagueId = req.params["id"];
+      const hasSeasons = await Season.findAll({
+        where: {
+          league_id: leagueId,
+        },
+      });
+
+      if (hasSeasons.length > 0) {
+        return res.status(200).json({
```
garcialuis, line 109: The appropriate status code in this case would most likely be [409 Conflict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/409) instead of 200 OK.

garcialuis (changes requested):
I reviewed the publish workflow and the server code changes only, left a comment on the changes to league controller.

PR #277, opened 2025-03-20
Bug bash/fix update functions

saulRM10:
### Checklist

### DIFF
--- a/server/controllers/division.controller.js
+++ b/server/controllers/division.controller.js
@@ -62,7 +62,11 @@ const DivisionController = {
 
     try {
       await modelByPk(res, Group, form.group_id);
-      const isUnique = await isDivisionNameUnique(form.group_id, form.name);
+      const isUnique = await isDivisionNameUnique(
+        form.group_id,
+        form.name,
+        false,
+      );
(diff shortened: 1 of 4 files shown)


server/controllers/division.controller.js
```diff
@@ -62,7 +63,11 @@ const DivisionController = {
 
     try {
       await modelByPk(res, Group, form.group_id);
-      const isUnique = await isDivisionNameUnique(form.group_id, form.name);
+      const isUnique = await isDivisionNameUnique(
+        form.group_id,
+        form.name,
+        false,
```
garcialuis, line 69: nit pick: Here instead of sending false when there's no divisionId I would send `null` instead. Even though it is not typed divisionId should be a number or null but in this case it's a number or a boolean.

garcialuis (approved):
Read through your changes, they LGTM. Thanks for updating those update functions to exclude self from the searches.

server/controllers/division.controller.js
saulRM10: good pint, updated for leagues, season and divisions

PR #279, opened 2025-03-21
Feature: Add Stripe Url for Payment Intents

saulRM10:
### Description

1. Added a `stripe_url` where it contains the link to the Payment Intent on the connected account's stripe dashboard .
2.  Added the `stripe_url` as a [virtual column](https://sequelize.org/docs/v6/core-concepts/getters-setters-virtuals/#virtual-fields) and used [`get()`](https://sequelize.org/docs/v6/core-concepts/getters-setters-virtuals/#getters) to set up the logic for the creation of the url. Shout out to @garcialuis for pointing me that direction. I think this is a WAY better way of updating query results with attributes we may need ! 
3. I did have to update the query to find the form `id` based of the form document id. For some reason we were using `findAll` but one `document_id` should only be linked to one form `id`.
4. Updated the front end to render a button to redirect users to stripe and see the transaction in more detail:
image

### Checklist

### DIFF
--- a/client/src/components/FormResponses/FormResponses.tsx
+++ b/client/src/components/FormResponses/FormResponses.tsx
@@ -93,6 +93,8 @@ const FormResponses = ({ formId, populateResponses }: FormResponsesProps) => {
   const [status, setStatus] = useState<("approved" | "rejected" | "pending")[]>(
     []
   );
+  const [stripeUrls, setStripeUrls] = useState<(string | null)[]>([]);
+
   const [formResponsesData, setFormResponsesData] = useState<AnswerRecordMap>(
     []
   );
@@ -167,6 +169,7 @@ const FormResponses = ({ formId, populateResponses }: FormResponsesProps) => {
       setPaymentIntentIds(paymentIntentIdsData);
       const statusData: ("approved" | "rejected" | "pending")[] = [];
       const paymentTypeData: string[] = [];
+      const stripeUrlsData: (string | null)[] = [];
       const formPayments = await getFormPayments(formId);
       await Promise.all(
         formPayments.map(
@@ -183,12 +186,14 @@ const FormResponses = ({ formId, populateResponses }: FormResponsesProps) => {
             } else {
               paymentTypeData[index] = "Cash / Check";
             }
+
+            stripeUrlsData[index] = paymentData.stripe_url || null;
           }
         )
       );
-
       setStatus(statusData);
       setPaymentType(paymentTypeData);
+      setStripeUrls(stripeUrlsData);
     })();
   }, [formId]);
 
@@ -379,6 +384,20 @@ const FormResponses = ({ formId, populateResponses }: FormResponsesProps) => {
                   </Modal.Button>
                   <Modal.Content maximize={true} title={user[index]}>
                     <FormResponseForm answers={response} />
+                    {stripeUrls[index] && (
+                      <PrimaryButton
+                        className={styles.stripePaymentButton}
+                        onClick={() => {
+                          if (stripeUrls[index]) {
+                            window.open(stripeUrls[index], "_blank");
+                          }
+                        }}
+                      >
+                        <p className={styles.btnTextDesktop}>
+                          View Stripe Payment
+                        </p>
+                      </PrimaryButton>
+                    )}
                   </Modal.Content>
                 </Modal>
               </td>
--- a/client/src/components/FormResponses/FormResponses.tsx
+++ b/client/src/components/FormResponses/FormResponses.tsx
@@ -93,7 +93,9 @@ const FormResponses = ({ formId, populateResponses }: FormResponsesProps) => {
   const [status, setStatus] = useState<("approved" | "rejected" | "pending")[]>(
     []
   );
-  const [stripeUrls, setStripeUrls] = useState<(string | null)[]>([]);
+  const [stripePaymentIntentUrls, setStripePaymentIntentUrlsData] = useState<
+    (string | null)[]
+  >([]);
 
   const [formResponsesData, setFormResponsesData] = useState<AnswerRecordMap>(
     []
@@ -169,7 +171,7 @@ const FormResponses = ({ formId, populateResponses }: FormResponsesProps) => {
       setPaymentIntentIds(paymentIntentIdsData);
       const statusData: ("approved" | "rejected" | "pending")[] = [];
       const paymentTypeData: string[] = [];
-      const stripeUrlsData: (string | null)[] = [];
+      const stripePaymentIntentUrlsData: (string | null)[] = [];
       const formPayments = await getFormPayments(formId);
       await Promise.all(
         formPayments.map(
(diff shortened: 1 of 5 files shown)


server/controllers/formPayment.controller.js
```diff
@@ -217,12 +217,16 @@ const FormPaymentController = function () {
         return res.status(formPayments.success).json(formPayments.error);
       }
 
-      const data = formPayments.data;
+      const data = formPayments.data.map((pyt) => {
+        return pyt.dataValues;
```
saulRM10, line 221: @garcialuis  I had to do this because a `findAll()` query returns a `object SequelizeInstance` and you can not update that directly (i am trying to add a new attribute `stripe_url` ). Just FYI, that took me a while to figure out that this was the reason my formPayments were not coming back with an updated `stripe_url` field

server/controllers/formPayment.controller.js
```diff
@@ -375,6 +378,21 @@ const FormPaymentController = function () {
...
   };
 
+  var addStripeUrl = function (formPayments) {
+    try {
+      formPayments.forEach((payment) => {
+        payment.stripe_url =
+          payment.payment_method_id === 1
+            ? `https://dashboard.stripe.com/payments/${payment.payment_intent_id}`
+            : null;
+        console.log("Modified payment:", JSON.stringify(payment));
```
garcialuis, line 388: console log left here, do you need it still?

garcialuis (changes requested):

server/controllers/formPayment.controller.js
garcialuis: Did you look into `setDataValue`? Here are the [docs](https://sequelize.org/docs/v7/models/getters-setters-virtuals/#setters). It may be useful to set the stripe_url to each of the `object SequelizeInstance`(s). And that would be done inside your `addStripeUrl` function since it's already iterating through every payment object.

server/controllers/formPayment.controller.js
garcialuis: Here's where I think we can use the `.setDataValue` to set the stripe_url.

server/controllers/formPayment.controller.js
saulRM10: Great feedback ! I incorporated a virtual column and used a get()

server/controllers/formPayment.controller.js
```diff
@@ -11,20 +11,19 @@ const FormPaymentController = function () {
...
-        });
-        formPayments = formPayments.concat(payments);
-      }
+      let payments = await FormPayment.findAll({
+        where: {
+          form_id: form.id,
+        },
+      });
+
+      formPayments = formPayments.concat(payments);
```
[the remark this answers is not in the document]
saulRM10: nice catch , updated !

client/src/components/FormResponses/FormResponses.tsx
```diff
@@ -183,12 +186,14 @@ const FormResponses = ({ formId, populateResponses }: FormResponsesProps) => {
             } else {
               paymentTypeData[index] = "Cash / Check";
             }
+
+            stripeUrlsData[index] = paymentData.stripe_url || null;
```
saulRM10, line 190: backed returns a field called `stripe_url` its either a string if the `formpayment` was an actual stripe transaction. It will be `null` if it was a cash payment.

client/src/components/FormResponses/FormResponses.tsx
```diff
@@ -379,6 +384,20 @@ const FormResponses = ({ formId, populateResponses }: FormResponsesProps) => {
                   </Modal.Button>
                   <Modal.Content maximize={true} title={user[index]}>
                     <FormResponseForm answers={response} />
+                    {stripeUrls[index] && (
+                      <PrimaryButton
+                        className={styles.stripePaymentButton}
+                        onClick={() => {
+                          if (stripeUrls[index]) {
```
saulRM10, line 391: image


I got a typescript error where there is an error when you try to use `null` as a URL. So i added a redundant check

client/src/components/FormResponses/FormResponses.tsx
```diff
@@ -93,6 +93,8 @@ const FormResponses = ({ formId, populateResponses }: FormResponsesProps) => {
   const [status, setStatus] = useState<("approved" | "rejected" | "pending")[]>(
     []
   );
+  const [stripeUrls, setStripeUrls] = useState<(string | null)[]>([]);
```
elimanzo, line 96: `stripeUrls` seem pretty vague. Are these invoices? If so, we can call it something like `stripeInvoiceUrls`

client/src/components/FormResponses/FormResponses.tsx
elimanzo: why not just change the type to (string | URL | undefined)[] and see if that works?

server/models/formpayment.js
```diff
@@ -102,6 +102,16 @@ module.exports = (sequelize, DataTypes) => {
         type: DataTypes.INTEGER,
         allowNull: true,
       },
+      stripe_url: {
+        type: DataTypes.VIRTUAL,
+        get() {
+          const paymentIntentId = this.payment_intent_id;
+          const paymentMethod = this.payment_method_id;
+          return paymentMethod === 1
```
elimanzo, line 110: `paymentMethod === 1` is really hard to understand as a reader what this means? I'm assuming this is if the `paymentMethod` is some payment type like `credit card | cash | etc..`

elimanzo (commented):
Sweet! I had couple of questions but this seems good to me!

client/src/components/FormResponses/FormResponses.tsx
jipatino98: Yeah it seems like the type is defined as `string | URL | undefined` so the best thing to do is what @elimanzo suggested and add `string | URL | undefined | null` to the type declaraiton. Better yet, if `stripeUrls` can be `undefined`, just omit the undefined in the type declaration and append a `?`.
So it would look like this in `FormResponses/types.ts`:
`stripe_url?: string | URL | null`

client/src/components/FormResponses/types.ts
```diff
@@ -66,6 +66,7 @@ export interface FormPaymentType {
   updated_at: Date;
   user_stripe_account_id: number;
   response_document_id: string | null;
+  stripe_url: string | null;
```
jipatino98, line 69: As stated in the other comment, it might be best to make this:
`stripe_url?: string | URL | null`

server/models/formpayment.js
jipatino98: To add onto this, maybe we can add a constant to top of the file to make this clearer. Maybe something like:
`const CREDIT_CARD = 1` to make it more readable.

jipatino98 (commented):
Just a few nits and suggestions before approval!

client/src/components/FormResponses/FormResponses.tsx
saulRM10: but the server is returning a string with a url

client/src/components/FormResponses/FormResponses.tsx
saulRM10: good point

client/src/components/FormResponses/types.ts
saulRM10: Issue is the server returns a `string` of a url, not a type of `URL`

garcialuis (approved):
updates to formPayment controller and model look good to me, thanks for the updates.

PR #280, opened 2025-03-22
update variables, should not be constants

garcialuis:
### Description
Changes some variables in the user controller, they were constants but in some cases a new value is being set. This PR changes those variables so that they are no longer constants. 

### Issue Link
Issue [Link](https://cascarita-org.sentry.io/issues/6441499871/?project=4508911777742848&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=1).

### DIFF
--- a/server/controllers/user.controller.js
+++ b/server/controllers/user.controller.js
@@ -103,7 +103,7 @@ const UserController = function () {
   };
 
   var registerUser = async function (req, res, next) {
-    const {
+    var {
       group_id,
       first_name,
       last_name,
@@ -120,8 +120,6 @@ const UserController = function () {
     const userBasicInfo = await getUserInfoFromAuth0(req.headers.authorization);
 
     let groupId = group_id;
-    // const groups = await GroupController.getGroupByName(name);
-    // groupId = groups[0].id;
 
     if (!groupId) {
       try {


server/controllers/user.controller.js
```diff
@@ -103,7 +103,7 @@ const UserController = function () {
   };
 
   var registerUser = async function (req, res, next) {
-    const {
+    var {
```
abanuelo, line 106: consider using `let`: https://sentry.io/answers/difference-between-let-and-var-in-javascript/

abanuelo (changes requested):

server/controllers/user.controller.js
garcialuis: @abanuelo These variables are defined at the beginning of the function and I do want them available throughout the whole function. What would be the benefit or using `let` vs `var` in this case?

abanuelo: @garcialuis No benefit at the moment but if we decide to reuse variable names scoped to different blocks later on, might be helpful. Will grant the approval for now but just wanted to point your attention to the distinction!

abanuelo (approved):

PR #281, opened 2025-03-24
fix: update publish.yml and create-env files

abanuelo:
### Description

Working through the publish workflow to be able to draft and generate new latest images with secrets fetched from Secrets Manager.

### Checklist

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -32,9 +32,7 @@ jobs:
         with:
           aws-access-key-id: ${{ secrets.AWS_ECR_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_ECR_SECRET_ACCESS_KEY }}
-          aws-region: us-west-1
-          profile-name: cascarita
-          config-file-path: ~/.aws/config
+          aws-region: ${{ secrets.AWS_REGION }}
 
       - name: Configure AWS profile
         run: |
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -33,19 +33,28 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ECR_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_ECR_SECRET_ACCESS_KEY }}
           aws-region: us-west-1
+          profile-name: cascarita
+          config-file-path: ~/.aws/config
+
+      - name: Configure AWS profile
+        run: |
+          aws configure --profile cascarita  set aws_access_key_id ${{ secrets.AWS_ECR_ACCESS_KEY_ID }}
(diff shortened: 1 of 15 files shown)


.github/workflows/publish.yml
```diff
@@ -33,6 +33,15 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ECR_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_ECR_SECRET_ACCESS_KEY }}
           aws-region: us-west-1
+          profile-name: cascarita
+          config-file-path: ~/.aws/config
```
garcialuis, line 37: @abanuelo I don't see (`profile-name` or `config-file-path`) as an [available option here](https://github.com/aws-actions/configure-aws-credentials/blob/main/README.md#options).

abanuelo: Good catch this was AI generated and part of previous commits that failed. Removed!

.github/workflows/publish.yml
```diff
@@ -32,7 +32,14 @@ jobs:
...
           aws-secret-access-key: ${{ secrets.AWS_ECR_SECRET_ACCESS_KEY }}
-          aws-region: us-west-1
+          aws-region: ${{ secrets.AWS_REGION }}
+
+      - name: Configure AWS profile
+        run: |
+          aws configure --profile cascarita  set aws_access_key_id ${{ secrets.AWS_ECR_ACCESS_KEY_ID }}
+          aws configure --profile cascarita  set aws_secret_access_key ${{ secrets.AWS_ECR_SECRET_ACCESS_KEY }}
+          aws configure --profile cascarita set region ${{ secrets.AWS_REGION }}
+          aws configure list
```
garcialuis, line 42: What is the purpose of setting up the profile here?

garcialuis (commented):
The changes themselves look good, can you remind me why we need to set up the aws profile after aws has already been configured?

.github/workflows/publish.yml
abanuelo: The action `aws-actions/configure-aws-credentials@v4` does not create an AWS profile on the GitHub Ubuntu Runner. In order for use to run the `create-env.js` script to fetch the secrets from Secrets Manager we need to have:
1. an aws config file (ie `~/.aws/config`) with a profile named `cascarita` 
2. an aws credentials file (ie `~/.aws/credentials`) with the access key id and the secret access key pointing to the `cascartia` profile

The reason why our publish workflow was failing before was I had the impression that the action we were using did this as a given but it does not. This new step was added to get the infra ready on the runner to be able to successfully fetch the secrets!

garcialuis: Thank you for the clarification, that makes sense. I think we had not gotten to this point in the past, that's why we didn't have the need to pull from secret manager. Thanks for adding this in!

garcialuis (approved):

PR #284, opened 2025-03-25
fix: adding teams query for player block

abanuelo:
### Description

Working to include teams query within player block. This prevents the need to re-create an entire form when new teams are created. See demo below

Recording 2025-03-25 at 15 15 04

### Checklist

### DIFF
--- a/client/src/components/FormInputComponents/PlayerBlock/PlayerBlock.tsx
+++ b/client/src/components/FormInputComponents/PlayerBlock/PlayerBlock.tsx
@@ -1,20 +1,31 @@
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { FieldProps } from "../types";
 import { FieldError, useFormContext } from "react-hook-form";
 import styles from "./PlayerBlock.module.css";
 import { useTranslation } from "react-i18next";
 import { ShortTeam } from "../../../api/forms/types";
+import { useGetTeamsBySeasonDivisionId } from "../../../api/teams/query";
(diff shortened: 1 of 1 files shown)


client/src/components/FormInputComponents/PlayerBlock/PlayerBlock.tsx
```diff
@@ -1,20 +1,31 @@
...
 
+  const seasonId = watch(`answers.${index}.player.season_id`);
+  const divisionId = watch(`answers.${index}.player.division_id`);
   const { required } = field.validations ?? {};
   const [isOther, setIsOther] = useState(false);
+  const { data: teams, refetch } = useGetTeamsBySeasonDivisionId(seasonId, divisionId);
+
+  useEffect(() => {
+    refetch();
+  }, [seasonId, divisionId, refetch]);
```
elimanzo, line 27: Do you think this is something we'll need to do every time we use the `useGetTeamsBySeasonDivisionId` hook, and if so should this live inside of it?

elimanzo (approved):
looks good! I had one small question but nothing major

client/src/components/FormInputComponents/PlayerBlock/PlayerBlock.tsx
elimanzo: edit: I just saw that 
```
const seasonId = watch(`answers.${index}.player.season_id`);
const divisionId = watch(`answers.${index}.player.division_id`);
```
ignore last comment 😄

PR #288, opened 2025-03-27
feat: adding prettier formatter for projects

elimanzo:
### Checklist


elimanzo:
* **#288** <a href="https://app.graphite.dev/github/pr/cascarita-io/cascarita/288?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/cascarita-io/cascarita/288?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

.prettierignore
```diff
@@ -0,0 +1,4 @@
+**/dist/**
+**/node_modules/**
+**/*.generated.js
```
elimanzo, line 3: if we have any generated files (usually migrations), we should append `.generated` to them

.prettierrc
```diff
@@ -5,9 +5,10 @@
...
   "singleQuote": false,
   "tabWidth": 2,
+  "endOfLine": "lf",
   "semi": true,
-  "importOrder": ["^react(.*)", "@/(.*)", "^[./]"],
+  "importOrder": ["^react(.*)", "<THIRD_PARTY_MODULES>", "^@/(.*)", "^[./]"],
   "importOrderSeparation": true,
   "importOrderSortSpecifiers": true,
-  "plugins": []
+  "plugins": ["@trivago/prettier-plugin-sort-imports"]
```
elimanzo, line 13: This was needed to run the import orders

image

Makefile
```diff
@@ -29,3 +29,7 @@ test:
 
 eslint:
 	cd client && npm run eslint
+
+format:
+	cd client && npm run format
+	cd server && npm run format
```
elimanzo, line 35: every project has a format script, but this allows you to format both

client/dist/assets/index-BXgSqAFy.js
elimanzo, line 1: 💀 prettier ignore didn't work here. Investigating

.gitignore
```diff
@@ -13,7 +13,7 @@ AWSCLIV2.pkg
 coverage
 
 # Build binaries not needed
-client/dist/*
+dist/
```
elimanzo, line 16: This wasn't properly ignoring our dist builds

client/.prettierignore
elimanzo, line 1: These had to be added in each project to properly ignore the files. There might be a way to implement a root level formatter, and linter, but that could be in a different project.

.prettierignore
saulRM10: we can try it but don't think its necessary. How can i append ? Using sequalize cli to auto generate the migration (and model incase the migration is creating a new table)

elimanzo: I see! If there's no customization of naming the files, then we can ignore the directory that has all the auto generated content. If you know all the files / directories that are auto generated, then we can add them in this PR or another 😄

saulRM10: auto generated files(form sequelize cli) are in stored in the server/migrations  , server/models and server/seeders

elimanzo: I was reading these files, are these files that are generated and never touched again or are they generated as a template? If it's the latter then we should just keep things as is, because we're editing them.

saulRM10: They are generated as a template

garcialuis (commented):
Changes to the server LGTM. I'll refer the final approval to Armando and Jose so that they get a chance to review the updates to the client. 
Thanks for getting this in Eli! 👍

PR #292, opened 2025-03-30
Fix/wrong group code

jipatino98:
### Description

This PR introduces a fix to a bug where when a user enters an incorrect group code when trying to join an existing group in registration modal, it sends them back to beginning of form, without any explanation. It also modularizes some ui elements in the `RegistrationModal.tsx` file into their own separate UI component. 

It also add form validation checking with React Hook Form rather than from the native `form` element.

### Issue Link

[Jira Ticket
](https://cascarita.atlassian.net/browse/CV-180)
### Screenshots

https://github.com/user-attachments/assets/973fa22e-6298-4619-8476-4471b7ee710c


client/src/api/users/types.ts
```diff
@@ -33,23 +19,60 @@ export interface UserSettingsResponse {
...
-  group_code: string | null;
-  token: string | null;
+export interface RegisterFormData {
+  has_accepted_tc: boolean;
+
+  first_name: string;
+  last_name: string;
+  language_id?: string;
+
+  is_existing_org: "yes" | "no";
```
elimanzo, line 29: I'd usually expect this to be a `boolean`. Is this the type that is needed to submit to the backend?

client/src/api/users/types.ts
```diff
@@ -33,23 +19,60 @@ export interface UserSettingsResponse {
...
-  zipCode: string | null;
-  logoUrl: string | null; // still need to set up s3 buckets for images so for now we wont collect this and just set it to null
-  group_code: string | null;
-  token: string | null;
+export interface RegisterFormData {
+  has_accepted_tc: boolean;
+
+  first_name: string;
+  last_name: string;
+  language_id?: string;
```
elimanzo, line 27: What does `language_id` mean? Is this supposed to take in a language from [these ](https://github.com/ladjs/i18n-locales?tab=readme-ov-file#usage)?

client/src/components/RegistrationModal/RegistrationModalPageFour.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+const RegistrationModalPageFour = () => {
+  const {
+    register,
+    formState: { errors },
+  } = useFormContext();
+
+  return (
+    <>
+      <div style={{ display: "grid", gap: "24px" }}>
```
elimanzo, line 13: I think the pattern we have here is to use `className` instead of inline styles

client/src/components/RegistrationModal/RegistrationModalPageFour.tsx
```diff
@@ -0,0 +1,104 @@
...
+          <input
+            {...register("org_name")}
+            className={styles.input}
+            id="orgName"
+            placeholder="Organization Name"
+            type="text"
+          />
+          {errors.org_name && (
+            <span className={styles.error}>
+              {String(errors.org_name.message)}
```
elimanzo, line 25: just a nit, but I rather have template strings like

```suggestion
               `{${errors.org_name.message}}`
```

client/src/components/RegistrationModal/RegistrationModalPageThree.tsx
```diff
@@ -0,0 +1,117 @@
...
+                  className={styles.label}
+                  htmlFor={`is_existing_org-${option.value}`}
+                >
+                  {option.label}
+                </label>
+              </div>
+            ))}
+          </fieldset>
+
+          {isExistingOrg === "yes" && (
```
elimanzo, line 53: same comment as above if we make this a boolean it'd be better to do just `isExistingOrg`. If we want to specifically display yes or no depending on the action, then we should handle that as a helper to show it.

client/src/components/RegistrationModal/RegistrationModalPageTwo.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { useFormContext } from "react-hook-form";
+
+const RegistrationModalPageTwo = () => {
+  const {
+    register,
+    formState: { errors },
+  } = useFormContext();
+
+  return (
+    <div style={{ display: "grid", gap: "24px" }}>
```
elimanzo, line 11: sane comment we'd want to use `className`, but looking back at this we need to create design system where we can just call `Grid` or `Flex` as a component to do stuff like this (In a different PR).

client/src/api/users/types.ts
```diff
@@ -33,23 +19,60 @@ export interface UserSettingsResponse {
...
-  name: string | null;
-  streetAddress: string | null;
-  city: string | null;
-  state: string | null;
-  zipCode: string | null;
-  logoUrl: string | null; // still need to set up s3 buckets for images so for now we wont collect this and just set it to null
-  group_code: string | null;
-  token: string | null;
+export interface RegisterFormData {
+  has_accepted_tc: boolean;
```
elimanzo, line 23: naming could be more explicit. How about `has_accepted_terms`?

client/src/components/RegistrationModal/RegistrationModalPageThree.tsx
```diff
@@ -0,0 +1,117 @@
...
+    label: "Yes",
+    value: "yes",
+  },
+  {
+    label: "No",
+    value: "no",
+  },
+];
+
+export default RegistrationModalPageThree;
```
elimanzo, line 117: A small nit, usually constants are at the top of the file after the imports

elimanzo (approved):
Couple of questions / nits, but accepting to unblock.

client/src/api/users/types.ts
jipatino98: One of the features we had to postpone for now is having the user select the language throughout the app. We had English and Spanish but due to time constraint, I removed the option to select from the frontend. The value of the language was represented by an Id, `number`. 
1 is for English
2 is for Spanish

client/src/api/users/types.ts
jipatino98: Since I used a radio input, the value should be a string, at least it is what i understood from the [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#value). However, I could make this a checkbox input instead and have it a boolean. I believe i choose a radio input simply because I phrased the prompt: _Are you joining an existing team?_

client/src/components/RegistrationModal/RegistrationModalPageFour.tsx
jipatino98: I was also thinking about using `className` but since I was only using two properties, I figured i wouldn't bloat the CSS file, I would so an inline style. However, i do like the idea you proposed about having a Flex and Grid component

client/src/components/RegistrationModal/RegistrationModalPageFour.tsx
jipatino98: I agree, I'll fix that and continue going forward

client/src/components/RegistrationModal/RegistrationModalPageThree.tsx
jipatino98: Would you think it would be better in the UI to have a checkbox instead? I don't think you are able to use the `checked` attribute to return a boolean value with radio inputs since it will always return true.

elimanzo: Yeah I think checkbox for `Yes` would be a better UI. I think having 2 options for Yes and No seems kind of redundant?

client/src/components/RegistrationModal/RegistrationModalPageThree.tsx
jipatino98: I'll make that change! I left it at the bottom as to not bloat the top of the file before the JSX. Maybe in the future, we can start creating `constants.js` files if we start reusing them.

elimanzo (approved):
LGTM! Thanks for addressing comments

PR #293, opened 2025-03-31
Feature:  get Player Data off Session id 

saulRM10:
### Description

This pr leverages the association on the `UserSessions` model by using [eager loading] (https://sequelize.org/docs/v6/advanced-association-concepts/eager-loading/#fetching-all-associated-elements). The article linked shows how we can use eager loading with sequelize query. In short eager loading means we get all the related data upfront. At the SQL level this is a query with one or more joins. 

### Checklist


server/controllers/user.controller.js
```diff
@@ -810,6 +812,73 @@ const UserController = function () {
     }
   };
 
+  var getPlayerDataByUserSessionId = async function (req, res) {
+    const id = req.params.user_session_id;
+    if (!id || isNaN(Number(id))) {
```
elimanzo, line 817: Nothing to change here but I really wish we used `UUIDs` :(

server/controllers/user.controller.js
```diff
@@ -810,6 +812,73 @@ const UserController = function () {
...
+      return res.status(400).json({
+        success: false,
+        error: `invalid value for id`,
+      });
+    }
+
+    try {
+      let data = await UserSessions.findOne({
```
elimanzo, line 825: nit data isn't changing so `const` is appropriate

```suggestion
      const data = await UserSessions.findOne({
```

server/controllers/user.controller.js
```diff
@@ -810,6 +812,73 @@ const UserController = function () {
...
+                    attributes: ["name"],
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+        raw: true,
+        nest: false,
+      });
```
elimanzo, line 866: This query looks so crazy I love it 💯

elimanzo (commented):
Nice! I learned a lot of about these query call. Also had a nit!

